PR #8, opened 2018-01-23
Placeholder text (and some linting)

mgerring:


server/models/ParameterSpec.py
```diff
@@ -56,7 +55,7 @@ class Meta:
     ui_only = models.BooleanField(default=False)            # Don't bother pushing value to server
     multiline = models.BooleanField(default=False)          # For edit fields
     derived_data = models.BooleanField(default=False)       # Don't save in the undo stack, it comes from other params
-
+    placeholder = models.TextField(null=True, blank=True)   # Placeholder/help text. Different from default in that it's not actually a value.
```
jstray: Remove null=True as we will use empty string to mean no placeholder

assets/js/WfParameter.js
```diff
@@ -156,6 +156,7 @@ export default class WfParameter extends React.Component {
               defaultValue={this.props.p.value}
               onBlur={this.blur}
               onKeyPress={this.keyPress}
+              placeholder={this.props.p.parameter_spec.placeholder || ''}
```
jstray: || '' not needed if we guarantee empty string from server

assets/js/WfParameter.js
```diff
@@ -173,6 +174,7 @@ export default class WfParameter extends React.Component {
...
```
jstray: here too

jstray:
Need tests:
- test_initmodules to ensure read correctly
- test_parameterval_views to ensure placeholder text comes out of API
- WfParameter test in placeholder case only (snapshots for text fields?)

server/tests/test_parameterval.py
```diff
@@ -16,7 +16,13 @@ def createTestWorkflow(self):
             id_name='stringparam',
             module_version=self.module_version,
             type= ParameterSpec.STRING,
-            def_value='foo')
+            def_value='foo',
+            placeholder='placeholder')
```
jstray: Later tests don't actually check that the placeholder text is set correctly, but then again it's not reasonable to test that ParameterSpec.objects.create() works correctly as Django tests already do that. So is this placeholder here unnecessary since we never check it? Probably not, as there could conceivably be an edge case where something goes wrong if the placeholder is set.

server/tests/test_parameterval_views.py
```diff
@@ -37,8 +37,17 @@ def test_parameterval_detail_get(self):
         self.assertEqual(str_val['parameter_spec']['name'], 'StringParam')
         self.assertEqual(str_val['parameter_spec']['id_name'], 'stringparam')
         self.assertEqual(str_val['parameter_spec']['type'], ParameterSpec.STRING)
+        self.assertEqual(str_val['parameter_spec']['placeholder'], 'placeholder')
```
jstray: I generally try to avoid using test text that is the same as other text. E.g. use something other than 'placeholder' for the placeholder text. This avoids both human confusion and certain potential bugs.

server/tests/test_parameterval.py
jstray: Oh wait, I get it -- this is the base class for the view tests as well.

server/tests/test_parameterval_views.py
```diff
@@ -37,8 +37,17 @@ def test_parameterval_detail_get(self):
...
         self.assertEqual(str_val['value'], 'fooval')
 
+        # parameters have correct types and values
```
jstray: Why the duplicated comment? It was introducing the block of parameter tests -- either remove or make it specific to this test you added, if you think capturing your intent here is useful.

PR #13, opened 2018-01-26
Count by date

mgerring:
Doing this here so we can see the diff from the old version, will move out to an external module afterwards

### DIFF
--- a/server/dispatch.py
+++ b/server/dispatch.py
@@ -171,7 +171,7 @@ def module_dispatch_output(wf_module, table, **kwargs):
     params = create_parameter_dict(wf_module, table)
     # got some error handling in here if, for some reason, someone tries to call
     # output on this and it doesn't have any defined html output
-    html_file = open(html_file_path, 'r+')
+    html_file = open(html_file_path, 'r+', encoding="utf-8")
     html_str = html_file.read()
 
     return (html_str, tableout, params)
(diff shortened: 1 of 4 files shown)


server/dispatch.py
```diff
@@ -171,7 +171,7 @@ def module_dispatch_output(wf_module, table, **kwargs):
     params = create_parameter_dict(wf_module, table)
     # got some error handling in here if, for some reason, someone tries to call
     # output on this and it doesn't have any defined html output
-    html_file = open(html_file_path, 'r+')
+    html_file = open(html_file_path, 'r+', encoding="utf-8")
```
jstray: Is this part of CountByDate?

server/tests/modules/test_countbydate.py
```diff
@@ -10,25 +10,61 @@ def setUp(self):
...
         # Use out.to_csv() instead of str(out) to ensure rows are output in index order (otherwise variable)
         set_string(self.col_pval, 'Date')
-        set_integer(self.sort_pval,0)  # 0 = sort by value
-        out = execute_wfmodule(self.wf_module)
-        self.assertEqual(out.to_csv(index=False), 'date,count\n2011-07-25,1\n2016-01-10,2\n' )
 
         # sort by freq
-        set_integer(self.sort_pval,1)  # 1 = sort by freq
+        set_integer(self.sort_pval, 1)  # 1 = sort by freq
+        out = execute_wfmodule(self.wf_module)
```
jstray: No need to do this now, but we could and should make countbydate and external module. One step in doing that would be to call render directly, like https://github.com/CJWorkbench/filter/blob/master/test_filter.py#L23

PR #14, opened 2018-01-26
Module Category closes on mouse out

Whelpley:


assets/js/ModuleCategory.js
```diff
@@ -35,6 +36,11 @@ export default class ModuleCategory extends React.Component {
     this.props.setOpenCategory(newCollapsed ? null : this.props.name); // tell parent, so it can close other cats
   }
 
+  collapseAll() {
+    this.setState({collapsed: true});
```
jstray: Why do we need to both set our state and the parent's state?

Whelpley: We don't! Setting the parent is sufficient, removed the setState from this both collapseAll() and toggleCollapse()

mgerring: Is it actually true that display:none is worse for performance than removing and then re-adding things to the DOM?

PR #15, opened 2018-01-26
Turn on alert on click

mgerring:
The branch name is a misnomer, it doesn't toggle. If alerts are turned off, alerts are turned on. If they're already on, nothing happens.


assets/js/AddNotificationButtonClosed.js
```diff
@@ -45,6 +47,20 @@ class AddNotificationButtonClosed extends React.Component {
     this.setState({showButton: false});
   }
 
+  handleClick() {
```
jstray: I'd already forgotten what this logic does and had to read it to figure out -- would be useful to have a short comment like "Add notification to the first module that adds data."

assets/js/AddNotificationButtonOpen.js
```diff
@@ -6,8 +6,9 @@
  */
 
 import React from 'react';
-import PropTypes from 'prop-types'
+import PropTypes from 'prop-types' //TODO: actually add prop types
```
jstray: Why not add them now? They are useful for detecting bugs.

assets/js/AddNotificationButtonOpen.test.js
```diff
@@ -10,17 +10,79 @@ import AddNotificationButtonOpen  from './AddNotificationButtonOpen'
...
+  var wrapper;
+  var notificationsOn;
+  var notificationsOff;
 
-  beforeEach(() => wrapper = mount(
-    <DragDropContextProvider backend={HTML5Backend}>
-      <AddNotificationButtonOpen/>
-    </DragDropContextProvider>
-  ));
+  beforeEach(() => {
```
jstray: This looks nearly identical to the closed test. Can it be factored out somewhere? Is it generic enough to become a standard test workflow in utils? If not even just sharing between these two tests is a good idea. Cut and paste code gives me hives.

mgerring:
Boy, that was an ugly merge

mgerring:
Manually merged

PR #18, opened 2018-01-30
Module Library collapse state persists after exiting workflow

Whelpley:


assets/js/ModuleLibrary.js
```diff
@@ -118,12 +118,16 @@ export default class ModuleLibrary extends React.Component {
   toggleLibrary() {
     if (!this.props.isReadOnly) {
       this.setState({ libraryOpen: !this.state.libraryOpen });
+      // make API call to set
```
jstray: A rare instance where I'd say the comment isn't helping much. General guideline: comment the why, not the what.

assets/js/ModuleLibrary.test.js
```diff
@@ -100,8 +112,8 @@ describe('ModuleLibrary', () => {
     
     it('Renders in closed state, without modules', () => {
       expect(wrapper).toMatchSnapshot();
-      // should NOT call getModules (one call from previous test)
-      expect(api.getModules.mock.calls.length).toBe(1);
+      // should NOT call getModules (two calls from previous tests)
+      expect(api.getModules.mock.calls.length).toBe(2);
```
jstray: This needs to be fixed. Tests should be independent of each other, so you can reorder them without breaking. Move the creation of getModules mock into beforeEach()  so this is always .toBe(1)

assets/js/WorkbenchAPI.js
```diff
@@ -3,6 +3,12 @@
 
 import { csrfToken } from './utils'
 
+const apiHeaders = {
```
jstray: Nice factoring.

server/tests/test_workflow_views.py
```diff
@@ -253,3 +254,35 @@ def test_workflow_title_post(self):
...
+    def test_workflow_public_post(self):
+        pk_workflow = Workflow.objects.get(name='Workflow 1').id
+        request = self.factory.post('/api/workflows/%d' % pk_workflow,
+                                   {'public': True})
+        force_authenticate(request, user=self.user)
+        response = workflow_detail(request, pk=pk_workflow)
+        self.assertIs(response.status_code, status.HTTP_204_NO_CONTENT)
+
+        # see that we get the new value back
+        request = self.factory.get('/api/wfmodules/%d/' % pk_workflow,)
```
jstray: I miscommunicated earlier; rather than calling get() this should just check the actual public field
```
workflow =  Workflow.objects.get(name='Workflow 1')
self.assertEqual(workflow.public, False)
pk_workflow = workflow.id
...
workflow.refresh_from_db()
self.assertEqual(workflow.public, True)
```

server/tests/test_workflow_views.py
```diff
@@ -253,3 +254,35 @@ def test_workflow_title_post(self):
...
+
+    def test_workflow_library_collapse_post(self):
+        pk_workflow = Workflow.objects.get(name='Workflow 1').id
+        request = self.factory.post('/api/workflows/%d' % pk_workflow,
+                                   {'module_library_collapsed': True})
+        force_authenticate(request, user=self.user)
+        response = workflow_detail(request, pk=pk_workflow)
+        self.assertIs(response.status_code, status.HTTP_204_NO_CONTENT)
+
+        # see that we get the new value back
```
jstray: Here too.

PR #20, opened 2018-01-31
Right pane resize

mgerring:
I know it doesn't have tests yet

### DIFF
--- a/assets/css/components/OutputPane.scss
+++ b/assets/css/components/OutputPane.scss
@@ -1,6 +1,7 @@
 /* Output Pane */
 .outputpane {
   flex: 1 1 auto;
+  position: relative;
 }
 
 .outputpane iframe {
--- a/assets/css/components/OutputPane.scss
+++ b/assets/css/components/OutputPane.scss
@@ -11,14 +11,13 @@
 }
 
 /* This div may be superfluous wrt .outputpane */
-.outputpane-box {
-  height: calc(100% - 6rem);
-  background-color: #F2F2F2;
-  position: relative;
-}
 
-iframe + .outputpane-box {
+iframe + .outputpane-table {
   height: calc(50% - 3rem);
+}
+.outputpane-table {
+  height:100%;
+  position:relative;
 }
 
 .outputpane-header {
@@ -34,4 +33,19 @@ iframe + .outputpane-box {
   left:0;
   right:0;
   bottom:0;
+}
+
+.overlapping .outputpane.focus .outputpane-box {
+  z-index: 2;
+  box-shadow: -2px 2px 5px rgba(0,0,0,0.3);
+}
+
+.overlapping .outputpane:not(.focus):after {
+  content:' ';
+  display: block;
+  top:0;
+  left:0;
+  right:0;
+  bottom:0;
+  position:absolute;
 }
--- a/assets/css/components/OutputPane.scss
+++ b/assets/css/components/OutputPane.scss
@@ -40,7 +40,11 @@ iframe + .outputpane-table {
   box-shadow: -2px 2px 5px rgba(0,0,0,0.3);
 }
 
-.overlapping .outputpane:not(.focus):after {
+/* Create an overlay to absorb click events so we don't change settings or set a different
+active wfmodule on focus change. We use the pseudo-element "after" to avoid writing an empty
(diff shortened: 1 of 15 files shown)


assets/css/components/OutputPane.scss
```diff
@@ -34,3 +34,18 @@ iframe + .outputpane-box {
   right:0;
   bottom:0;
 }
+
+.overlapping .outputpane.focus .outputpane-box {
+  z-index: 2;
+  box-shadow: -2px 2px 5px rgba(0,0,0,0.3);
+}
+
+.overlapping .outputpane:not(.focus):after {
```
jstray: :after is unusual and I don't think I understand what this CSS rule is doing. Can you add a comment?

assets/css/style.scss
```diff
@@ -1430,6 +1430,22 @@ img.logo {
...
   height:100vh;
+  position: relative;
+}
+.overlapping .modulestack.focus {
+  z-index: 2;
+  box-shadow: 2px 2px 5px rgba(0,0,0,0.3);
+}
+/* Create an overlay to absorb click events so we don't change settings or set a different
+active wfmodule on focus change */
+.overlapping .modulestack:not(.focus):after {
```
jstray: Ok, this comment helps, but why :after?

assets/js/OutputPane.js
```diff
@@ -137,6 +157,83 @@ export default class OutputPane extends React.Component {
...
+      let width = parseFloat(this.state.width) + ((d.width / this.state.pctBase) * 100) + '%';
+      this.setState({
+          initLeftOffset: this.state.leftOffset,
+          width,
+          resizing: false
+      });
+      this.props.setOverlapping((this.state.leftOffset < 0));
+  }
+
+  reset(libraryState, libraryToggle) {
```
jstray: reset() may not be the greatest method name -- what is being reset? Relatedly, a comment here wouldn't hurt.

assets/js/TableView.js
```diff
@@ -104,7 +108,8 @@ export default class TableView extends React.Component {
         rowGetter={this.getRow}
         rowsCount={this.props.totalRows }
         minWidth={this.state.gridWidth -2}
-        minHeight={this.state.gridHeight-2} />;   // -1 because grid has borders, don't want to expand flex grid
+        minHeight={this.state.gridHeight-2}
+        onRowClick={this.props.onClick} />;   // -1 because grid has borders, don't want to expand flex grid
```
jstray: This seems not to be passed in where TableView is created in OutputPane -- still needed? If you want to keep it, then, onClick should be in propTypes.

assets/js/WorkflowList.js
```diff
@@ -0,0 +1,180 @@
+import React from 'react'
```
jstray: I love that you split this out into a separate file. I have some reservations about the name. We use "Workflow list" elsewhere (e.g. https://github.com/CJWorkbench/cjworkbench/blob/master/assets/css/style.scss#L1251) to refer to the /workflows page, that is, the list of the user's workflows. The list of modules within a Workflow we have called ModuleStack (e.g. https://github.com/CJWorkbench/cjworkbench/blob/master/assets/css/style.scss#L1313). So we need to either change the name elsewhere or here.

assets/js/workflow.js
```diff
@@ -206,94 +30,93 @@ class Workflow extends React.Component {
...
+  }
+
   render() {
     // Wait until we have a workflow to render
     if (this.props.workflow === undefined) {
       return null;
     }
 
-    let selected_workflow_module_ref = this.props.workflow.wf_modules.find((wf) => {
+    let selectedWorkflowModuleRef = this.props.workflow.wf_modules.find((wf) => {
```
jstray: Yasss slowly getting toward a unified naming scheme.

assets/js/workflow.js
```diff
@@ -195,8 +14,13 @@ class Workflow extends React.Component {
   constructor(props: iProps) {
     super(props);
     this.state = {
-      isPublic: false
+        isPublic: false,
+        focus: false,
+        overlapping: false,
```
jstray: This would be a good place for a comment defining what "overlapping" means in all the code below

PR #22, opened 2018-01-31
Kb urls

mgerring:


server/models/Module.py
```diff
@@ -28,13 +28,16 @@ class Meta:
...
+    link = models.URLField('link', max_length=200, default="")
 
     # icon name associated with module
     icon = models.CharField('icon', max_length=20, default='url')
 
     # Does this model bring in external data?
     loads_data = models.BooleanField('loads_data', default=False)
 
+    # URL for the module's documentation, defaults to our knowledge base root
+    help_url = models.URLField('help_url', max_length=200, default=False)
```
jstray: What is default=False, and why False here and '' for link field?
Also, I suspect we want only partial (relative)  URLS for the help_url (see below), will URLField still handle that?

server/serializers.py
```diff
@@ -32,9 +33,14 @@ class Meta:
 
 
 class ModuleSerializer(serializers.ModelSerializer):
+    help_url = serializers.SerializerMethodField()
+
+    def get_help_url(self, obj):
+        return obj.help_url if obj.help_url else KB_ROOT_URL
```
jstray: Don't we have to add KB_ROOT_URL to help_url? I am envisioning that the URL in the json will be relative, so just 'add-data/add-from-url'. The idea is that we can seamlessly rebase the KB links, if e.g. we move off Intercom

server/serializers.py
```diff
@@ -36,7 +37,13 @@ class ModuleSerializer(serializers.ModelSerializer):
     help_url = serializers.SerializerMethodField()
 
     def get_help_url(self, obj):
-        return obj.help_url if obj.help_url else KB_ROOT_URL
+        url_pattern = re.compile('^http(?:s?)://', re.IGNORECASE)
```
jstray: btw you can just re.match() for a one-off, no need to compile

server/serializers.py
```diff
@@ -36,7 +37,13 @@ class ModuleSerializer(serializers.ModelSerializer):
...
+
+        if re.search(url_pattern, obj.help_url):
+            return obj.help_url
+
+        return f"{KB_ROOT_URL}{obj.help_url}"
```
jstray: I had to look up 'f'. Why not just +?

mgerring: didn't realize you could do that, I haven't dug in to the full breadth of python 3 format strings

PR #30, opened 2018-02-22
Library debugging

Whelpley:


assets/js/ModuleCategory.js
```diff
@@ -47,15 +52,13 @@ export default class ModuleCategory extends React.Component {
 
     var sortIcon = isOpen ? 'icon-sort-up-vl-gray' : 'icon-sort-down-vl-gray';
 
-    // ??? are these the correct icons per category ???
     const icons = {
       'Add data': 'database',
-      'Analyze': 'notepad',
+      'Analyse': 'notepad',         // Inconsistency with 'Analyze' elsewhere
```
jstray: Didn't we get rid of 'Analyse' category?

assets/js/ModuleCategory.js
```diff
@@ -47,15 +52,13 @@ export default class ModuleCategory extends React.Component {
...
       'Prepare':'wrangle',
       'Visualize': 'chart',
-      'Code': 'code',
+      'Code': 'code',               // not an active icon, needs replacement
```
jstray: What do you mean "needs replacement"?

assets/js/ModuleCategory.test.js
```diff
@@ -134,48 +201,74 @@ describe('ModuleCategory ', () => {
       // mouse enters category
       category.simulate('mouseEnter');
       // check: setOpenCategory() called from props
-      expect(setOpenCategory.mock.calls.length).toBe(3); // called twice before
+      expect(setOpenCategory.mock.calls.length).toBe(1);
```
jstray: much better

assets/js/ModuleCategory.test.js
```diff
@@ -134,48 +201,74 @@ describe('ModuleCategory ', () => {
...
+            isReadOnly={false}
+            collapsed={false}
+            setOpenCategory={setOpenCategory}
+            libraryOpen={false}
+          />
+        </DragDropContextProvider>
+      )
+    });
+    afterEach(() => {
+      wrapper.unmount();
```
jstray: may not be necessary, test?

assets/js/ModuleCategory.js
Whelpley: I was not aware of that, some modules still load into that category, however.

Should I search & replace any instances of "Analyse" with "Analyse"?

assets/js/WfModule.js
```diff
@@ -336,6 +336,13 @@ class WfModule extends React.Component {
                 />
               </div>
 
+    var helpIcon = undefined;
```
jstray: `var helpicon` would also set to undefined. If you intend no value, generally `var helpicon=null` is the way to go.

assets/js/WfModuleContextMenu.js
```diff
@@ -133,7 +131,7 @@ export default class WfModuleContextMenu extends React.Component {
 
     return (
       <Modal isOpen={this.state.exportModalOpen} toggle={this.toggleExportModal} className={this.props.className}>
-        <ModalHeader toggle={this.toggleModal} className='dialog-header modal-header d-flex align-items-center' >
+        <ModalHeader className='dialog-header modal-header d-flex align-items-center' >
```
jstray: Why remove toggleModal? What bug is this?

assets/js/ModuleCategory.js
Whelpley: Slight mixup there: I meant that there is no icon matching 'icon-code' in the CSS.

However, we do have the icon for 'code', but not for "Analyze/Analyse". (it looks like there are Modules in both categories)

assets/js/ModuleCategory.test.js
Whelpley: This is necessary because Jest can not have multiple HTML5 backends (due to wrapping the modules in a DragDropContextProvider)

assets/js/WfModule.js
Whelpley: Noted, will change where I find this elsewhere as well.

assets/js/WfModuleContextMenu.js
Whelpley: We have our own custom close icon in the modal:

<div className='icon-close' onClick={this.toggleExportModal}></div>

We could instead use `toggle={this.toggleModal}` in the ModalHeader, but we would be using the default 'X' icon to close, instead of our custom one.

assets/js/ModuleCategory.js
Whelpley: Have removed the deprecated module under the 'Analyse' category, all looks good now.

PR #31, opened 2018-02-22
Reduxify

mgerring:


mgerring:
Two of the test suites in npm aren't passing yet and there may be some issues with the python tests as well, but you can start reviewing now

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
 }
+registerReducerFunc(ADD_MODULE + '_FULFILLED', (state, action) => {
+  let insertBefore = action.payload.insert_before;
 
-// Call delete API, then dispatch a reload
-export function removeModuleAction(id_to_delete) {
+  delete action.payload.insert_before;
+
+  return update(state, {
+    selected_wf_module: {$set: action.payload.id},
```
jstray: Is this necessary, since you also dispatch setSelectedWfModuleAction? Maybe you can avoid that setSelectedWfModuleAction entirely, and hence the thunk.

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
 
-export function changeSelectedWfModuleAction(wfModuleID) {
-  return {
-    type : SELECTED_MODULE_CHANGE,
-    id : wfModuleID,
+  return function (dispatch) {
+    return (
+      // Set the new selected module before deleting to avoid errors.
+      // We do this even if the selected module isn't changing to avoid
+      // writing a subtly tricky conditional here.
```
jstray: It does cost an extra server round-trip, however. Or maybe we do this optimistically anyway so we delete on the client without waiting?

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+  let workflowID = store.getState().workflow.id;
   return {
-    type : TOGGLE_MODULE_COLLAPSED,
-    id : wfModuleID,
-    isCollapsed
+    type : SET_SELECTED_MODULE,
+    payload : {
+      promise: api.setSelectedWfModule(workflowID, wfModuleID).then((response) => {
+        return {
+          wf_module_id: response.selected_wf_module
```
jstray: Why is this needed? I would think this value only appears in _FULFILLED, and we don't bother handling it.

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+        }
+      }),
+      data: {
+        wf_module_id: wfModuleID
+      }
+    }
   }
 }
+registerReducerFunc(SET_SELECTED_MODULE + '_PENDING', (state, action) => {
+  if (!('selected_wf_module' in state) || (action.payload.wf_module_id !== state.selected_wf_module)) {
```
jstray: When would it happen that state does not have selected_wf_module, if it comes in from serializer?

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+// Patch a workflow module with new data
 
-            // Clone the state, switch out this one wfm (keep position in wf_modules array )
-            newState = Object.assign({}, state);
-            newState.workflow = Object.assign({}, state.workflow);
-            newState.workflow.wf_modules = state.workflow.wf_modules.map(el => el!==wfm ? el : newWfm);
-          }
+// TODO: We don't validate which fields or types are on
+// a WfModule here. The backend will reject nonexistent
+// fields, but should we do something on the frontend?
```
jstray: I'm comfortable not checking, for the following reason: the backend MUST check, as a security concern. If the front end calls this with bogus fields it's a bug, but no harm done. However it does bring up the issue of run time assertions. We currently don't have any mechanism for them, and it's not clear what we would do with assertion failures if we did.

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+  };
+}
+registerReducerFunc(UPDATE_WF_MODULE + '_PENDING', (state, action) => {
+  let moduleIdx = findIdxByProp(
+    state.workflow.wf_modules,
+    'id',
+    action.payload.id
+  );
+
+  if (typeof moduleIdx !== 'undefined') {
```
jstray: Who would call with undefined?

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+  return {
+    type : SET_WF_MODULE_STATUS,
+    payload: {
+      id : wfModuleID,
+      status : status,
+      error_msg: error_msg
+    }
+  }
+}
+registerReducerFunc(SET_WF_MODULE_STATUS, (state, action) => {
```
jstray: Don't we need  + "PENDING" here? Is this a bug? If so, do the tests catch it?

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
+}
+registerReducerFunc(MARK_DATA_VERSIONS_READ + '_PENDING', (state, action) => {
+  let wfModuleIdx = findIdxByProp(state.workflow.wf_modules, 'id', action.payload.id);
+  if (typeof wfModuleIdx !== 'undefined' &&
+    typeof state.workflow.wf_modules[wfModuleIdx].versions !== 'undefined') {
+    return update(state, {
+      workflow: {
+        wf_modules: {
+          [wfModuleIdx]: {
+            // Take the versions array,
```
jstray: good comments

assets/js/workflow-reducer.js
```diff
@@ -25,312 +41,567 @@ export function mockAPI(mock_api) {
...
-        return Object.assign({}, state, {
-          loggedInUser: action.user,
-        });
-      } else {
-        return state;
+export function clearNotificationsAction(wfModuleId) {
+  return {
+    type: CLEAR_NOTIFICATIONS,
+    payload: {
+      promise: api.deleteWfModuleNotifications(wfModuleId),
```
jstray: I wonder if this api could be merged into updateWfModule

assets/js/workflow-reducer.test.js
```diff
@@ -27,62 +45,216 @@ describe('Reducer actions', () => {
   };
 
   // Stub result to be returned by our stub loadWorkflow
-  const mock_load_workflow_result = { id: 1001 };
+  //const mock_load_workflow_result = { id: 1001 };
```
jstray: Delete if un-needed

assets/js/workflow-reducer.test.js
```diff
@@ -27,62 +45,216 @@ describe('Reducer actions', () => {
...
+  it('Adds a module', () => {
+    let state = workflowReducer(test_state, {
+      type: 'ADD_MODULE_FULFILLED',
+      payload: mock_add_wf_module_result,
+    });
+    expect(state.workflow.wf_modules[2].id).toEqual(40);
+    expect(state.selected_wf_module).toEqual(40);
+  });
+
+  it('Deletes a module', () => {
```
jstray: Would be good to have tests for boundary conditions, so that all code paths are exercised, e.g. delete last remaining module, delete currently selected module

assets/js/workflow-reducer.test.js
```diff
@@ -27,62 +45,216 @@ describe('Reducer actions', () => {
...
+        }
+      }
+    });
+    expect(state).toBe(test_state);
+  });
 
-    var actionPromise = removeModuleAction(30);
+  it('Sets the wfModule status', () => {
+    let state = workflowReducer(test_state, {
+      type: 'SET_WF_MODULE_STATUS',
```
jstray: I believe this should be + '_PENDING'

assets/js/workflow.page.js
```diff
@@ -27,16 +27,17 @@ const mapStateToProps = (state) => {
...
 const mapDispatchToProps = (dispatch) => {
   return {
     addModule: (module_id, insertBefore) => {
       dispatch(Actions.addModuleAction(module_id, insertBefore))
     },
     removeModule: (wf_module_id) => {
-      dispatch(Actions.removeModuleAction(wf_module_id))
+      dispatch(Actions.deleteModuleAction(wf_module_id))
     },
+    //TODO: make this use the redux action
```
jstray: Why not now?

server/tests/test_workflow_views.py
```diff
@@ -142,7 +142,10 @@ def test_workflow_addmodule_put(self):
         self.assertEqual(response.data['id'], wfm1.id)
         # we should get a full serialization back, same as /wfmodules/xx call
         serializer = WfModuleSerializer(wfm1)
-        self.assertEqual(response.data, serializer.data)
+        # addmodule serialization should also return insert_before property
+        return_data = serializer.data
+        return_data['insert_before'] = '0'
```
jstray: Could return order instead

server/serializers.py
```diff
@@ -87,11 +87,17 @@ def get_notification_count(self, wfm):
     def get_html_output(self, wfm):
         return wfm.module_version.html_output
 
+    versions = serializers.SerializerMethodField()
+    def get_versions(self, wfm):
```
jstray: This needs a test within test_wfmodule.py

assets/js/wfparameters/DataVersionSelect.js
```diff
@@ -39,36 +39,30 @@ export default class DataVersionSelect extends React.Component {
...
-
       if (!this.state.modalOpen === false) {
-        var idx = this.state.versions.versions.map((vers) => {
-          return vers[0]
-        }).indexOf(this.state.versions.selected);
-
+        // TODO: this.props.versions.selected should be a tuple of (<Date>, <Boolean>)
+        // TODO: Don't do Array.find, it's slow
+        let idx;
+        for (let i = 0; i < this.props.versions.versions.length; i++) {
```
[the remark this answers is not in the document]
mgerring: array.find is indefensibly slower than a loop. A loop is something like 90% faster. I can set up the jsperf test again if you want to see for yourself.

assets/js/workflow-reducer.js
mgerring: It's necessary to make the UI feel fast, yes, but it's not technically necessary. Without it there's a perceptible lag.

assets/js/wfparameters/DataVersionSelect.js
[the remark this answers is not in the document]
jstray: Ok. Can we put our loop in a fn for clarity and DRY?

assets/js/workflow-reducer.js
jstray: Is there some way to dispatch both at once? I see no technical problem having two promises unresolved at the same time, and the server-side race condition is irrelevant here

assets/js/wfparameters/DataVersionSelect.js
```diff
@@ -39,36 +39,30 @@ export default class DataVersionSelect extends React.Component {
...
+          idx = i;
+          if (this.props.versions.selected === this.props.versions.versions[i][0]) {
+            break;
+          }
+        }
         // If the selected version is false,
-        if (!this.state.versions.versions[idx][1]) {
```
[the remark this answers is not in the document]
mgerring: Nope, the comment still applies

assets/js/wfparameters/DataVersionSelect.js
```diff
@@ -216,6 +196,34 @@ export default class DataVersionSelect extends React.Component {
   }
 }
 
+const mapStateToProps = (state, ownProps) => {
+  let wfModuleIdx;
+  for (let i = 0; i < state.workflow.wf_modules.length; i++) {
```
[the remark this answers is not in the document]
mgerring: This is arguably a huge potential performance problem since this runs every time the state updates

assets/js/workflow-reducer.js
mgerring: No, there isn't. You can't dispatch setSelectedWfModuleAction until you have a wfModuleId, which you only have once the wfModule returns from the database.

assets/js/workflow-reducer.js
mgerring: No idea, that condition was there when I started working on this

mgerring: Actually, does it get returned from the serializer if it's a new workflow?

assets/js/workflow-reducer.js
mgerring: There's some weird cases where this could happen -- let's say a module is in some kind of queue to get updated (say, if the upstream data changed), and each update takes a long time, and before we get to this module, it gets deleted.

PR #33, opened 2018-02-26
Django allauth

mgerring:
There's an _account folder in templates containing templates we can optionally customize. We can delete it, but it's useful to have around.


cjworkbench/settings.py
```diff
@@ -149,8 +152,8 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'account.middleware.LocaleMiddleware',
-    'account.middleware.TimezoneMiddleware',
+    #'account.middleware.LocaleMiddleware',
```
jstray: Delete if no longer needed

cjworkbench/settings.py
```diff
@@ -167,7 +170,7 @@
                 'django.template.context_processors.request',
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
-                'account.context_processors.account',
+                #'account.context_processors.account',
```
jstray: this too

cjworkbench/urls.py
```diff
@@ -20,21 +20,21 @@
...
+#from cjworkbench.views.login import LoginView
 from cjworkbench.views.user import current_user, delete_google_creds
 from cjworkbench.google_oauth import authorize, get_creds
 
 urlpatterns = [
     url(r'^admin/?', admin.site.urls),
-    url(r'^xyzzy/signup/$', SignupView.as_view(), name='account_signup'),
-    url(r'^account/signup/$', page_not_found,  {'exception': Http404()}),
-    url(r'^account/login/$', LoginView.as_view(), name='account_login'),
+    #url(r'^xyzzy/signup/$', SignupView.as_view(), name='account_signup'),
```
jstray: don't we still need "secret" signup link?

integrationtests/test_login.py
```diff
@@ -19,11 +39,11 @@ def tearDownClass(cls):
     def test_login(self):
         b = self.browser
         b.visit(self.live_server_url + '/account/login')
-        b.fill('email', self.user.email)
+        b.fill('login', self.user.email)
         b.fill('password', self.password)
 
         b.find_by_tag('button').click()
-
+        time.sleep(2)
```
jstray: why does login now require a sleep when it did not before? Is it substantially slower?

templates/_account/email.html
```diff
@@ -0,0 +1,73 @@
+{% extends "account/base.html" %}
```
jstray: When does this template get displayed?

templates/_account/email/email_confirmation_message.txt
```diff
@@ -0,0 +1,8 @@
+{% load account %}{% user_display user as user_display %}{% load i18n %}{% autoescape off %}{% blocktrans with site_name=current_site.name site_domain=current_site.domain %}Hello from {{ site_name }}!
+
+You're receiving this e-mail because user {{ user_display }} has given yours as an e-mail address to connect their account.
```
jstray: typo "yours" should be "you"

mgerring: No, this is correct, "has given yours" where the possessive refers to the email address. 

Could be written as "you're receiving this e-mail because {{ user_display }} wants to connect this e-mail address to their account."

integrationtests/test_login.py
mgerring: I couldn't get this test to run locally without some kind of pause.

templates/_account/email.html
mgerring: It doesn't in our code, see note above.

cjworkbench/management/commands/load_socialaccounts.py
```diff
@@ -0,0 +1,38 @@
+# Startup code
+from django.core.management.base import BaseCommand, CommandError
+from django.contrib.sites.models import Site
+from cjworkbench import settings
+
+class Command(BaseCommand):
+    help = 'Loads the social accounts (with secrets) from the config file'
```
jstray: If I saw this message or found this file I don't think I would connect "social accounts" with login options. Can you add a brief comment to describe what we are doing here, and why it's a manage command (your 
 hard won knowledge)?

start-prod.sh
```diff
@@ -15,5 +15,5 @@ export PYTHONUNBUFFERED=0
 cron
 
 python manage.py migrate
+python manage.py load_socialaccount
```
jstray: Where the does the manage command for social accts run when testing on travis?
Also what happens if this command is not run? For example, on a dev install? Will app signup still work? Will there be any sort of error that suggests to the developer what the problem is?

mgerring: It doesn't run at all during tests -- we manually create the social accounts in the tests. The only way you would get an error in dev is if you a) manually create or download a socialaccounts_secrets.json file and then b) do not run the management command to load those social accounts into the database. Since a) is a manual process, as is /api/initmodules, I thought it would be reasonable for b) to be manual as well. I can update the wiki.

jstray: Yes please update the wiki. Do the signup and login screens still work if no secrets are installed?

mgerring:
Yes

On Feb 28, 2018 14:20, "Jonathan Stray" <notifications@github.com> wrote:

*@jstray* commented on this pull request.
------------------------------

In start-prod.sh
<https://github.com/CJWorkbench/cjworkbench/pull/33#discussion_r171404521>:

> @@ -15,5 +15,5 @@ export PYTHONUNBUFFERED=0
 cron

 python manage.py migrate
+python manage.py load_socialaccount

Yes please update the wiki. Do the signup and login screens still work if
no secrets are installed?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/CJWorkbench/cjworkbench/pull/33#discussion_r171404521>,
or mute the thread

PR #47, opened 2018-03-07
Embeds

mgerring:


assets/js/OutputIframe.js
```diff
@@ -1,28 +1,120 @@
...
+    }
+  }
+
+  setPublic() {
+    store.dispatch(setWorkflowPublicAction(this.props.workflow.id, true));
+  }
+
+  renderSetPublicModal() {
+    return (
+      <Modal isOpen={this.state.setPublicModalOpen} toggle={this.toggleModals} className='test-setpublic-modal'>
```
jstray: Is this the same code as the public modal in (IIRC) WorkflowMetadata? Should be a shared component if so.

assets/js/OutputIframe.js
```diff
@@ -1,28 +1,120 @@
...
+        {this.renderSetPublicModal()}
+        {this.renderEmbedCodeModal()}
+        <div className="outputpane-iframe-control-overlay">
+          <div className="btn icon icon-code" onClick={this.toggleModal} />
+        </div>
+        <iframe ref={this.setRef} src={this.state.src} />
+      </div>
+    );
   }
 }
```
jstray: Needs PropTypes

assets/js/workflow-reducer.js
```diff
@@ -122,6 +123,26 @@ registerReducerFunc(RELOAD_WORKFLOW + '_FULFILLED', (state, action) => {
...
+  return function (dispatch) {
+    return (
+
+      dispatch({
+        type: SET_WORKFLOW_PUBLIC,
+        payload: api.setWorkflowPublic(workflowId, isPublic)
+      }).then( () => {
+
+      dispatch(
+        reloadWorkflowAction()
```
jstray: Why is the reload necessary? Isn't the public flag in the redux store, and therefore ultimately passed as a prop to the navbar where it's displayed?

server/serializers.py
```diff
@@ -130,14 +130,14 @@ class WorkflowSerializer(serializers.ModelSerializer):
     owner_name = serializers.SerializerMethodField()
 
     def get_read_only(self, obj):
-        return obj.read_only(self.context['user'])
+        return obj.read_only(self.context.get('user', False))
```
jstray: Why is the False required? Could use a comment

server/views/embed.py
```diff
@@ -0,0 +1,27 @@
+from django.template.response import TemplateResponse
+from rest_framework.renderers import JSONRenderer
+from server.models import WfModule
+from server.serializers import WorkflowSerializer, WfModuleSerializer
+
+def embed(request, wfmodule_id):
+    try:
```
jstray: this can be shortened to get_object_or_404

templates/base.html
```diff
@@ -1,7 +1,7 @@
 {% load analytics_tags %}
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" {% block html_tag_attrs %}{% endblock %}>
```
jstray: Why is this needed?

assets/js/workflow-reducer.js
mgerring: It may not be necessary, but we have to deal with both 'public' and 'isReadOnly', and the way that both 'public' and 'readOnly' properties are handled needs a full re-write that I wasn't ready to deal with here.

server/serializers.py
mgerring: In case we have a request with no user, will add a comment

server/views/embed.py
mgerring: No it can't for two reasons: 1) Django doesn't support per-view 404 pages using get_object_or_404, if you want a custom 404 you have to do it manually. Also, this doesn't send a 404. It sets wf_module and workflow to 'none' and renders the same template, so we can handle both botched embed codes and embed codes that used to work but have been made private.

templates/base.html
mgerring: Can take this out now

assets/js/OutputIframe.js
mgerring: Sort of, but that works because WorkflowMetadata and NavBar share state. The entire handling of public and readOnly needs refactoring using redux. Then we can make this shared code.

templates/base.html
mgerring: I take that back, this is necessary to get the document height to 100% to make sure the iframe is sized correctly.

PR #49, opened 2018-03-07
Bug fix: text fields in WfParameter no longer drag on selection

Whelpley:


assets/js/WfParameter.js
```diff
@@ -308,7 +308,10 @@ export default class WfParameter extends React.Component {
         return (
           <div className='parameter-margin'>
             <div className='label-margin t-d-gray content-3'>{this.name}</div>
+            {/* need to disable dragsource on this */}
             <textarea
+              onMouseEnter={() => this.props.toggleDrag() }
```
jstray: Using toggle is ok for enter and leave is, but the drag state could get out of sync. More reliable would be setDrag(true) / setDrag(false)

PR #50, opened 2018-03-15
Add scroll position logic and animations to scroll on add/delete/focus

mgerring:


assets/js/WfModule.test.js
```diff
@@ -1,41 +1,39 @@
...
+  });
 
   it('Renders - dummy test only', () => {
-
-    expect(true).toBe(true);
-
-    // expect(wrapper).toMatchSnapshot();
-
-
+    expect(wrapper).toMatchSnapshot();
```
jstray: Set focusModule to jest.fn(), use mount() to get lifecycle methods called, check that setFocus was called once

PR #52, opened 2018-03-15
debounce click (and call on the leading edge of the timeout)

mgerring:
... to prevent doubleclick

### DIFF
--- a/assets/js/Module.js
+++ b/assets/js/Module.js
@@ -1,9 +1,9 @@
 /**
  * A component that holds a single module that is then contained within the
  * Module Library.
- * 
+ *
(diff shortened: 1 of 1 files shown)


assets/js/Module.js
```diff
@@ -41,7 +42,11 @@ function collect(connect, monitor) {
 class Module extends React.Component {
   constructor(props) {
     super(props);
-    this.itemClick = this.itemClick.bind(this);
+    this.itemClick = debounce(this.itemClick.bind(this), 1000, {
```
jstray: Logic looks fine. I’d like the comment to  why we are doing this  (Suppress multiple adds and dbl click)

PR #53, opened 2018-03-21
Select wf parameter text

Whelpley:


assets/js/WfParameter.js
```diff
@@ -409,5 +413,6 @@ WfParameter.propTypes = {
   changeParam:      PropTypes.func.isRequired,
 	getParamText:     PropTypes.func.isRequired,
   setParamText:     PropTypes.func.isRequired,
-	toggleDrag:       PropTypes.func.isRequired  
+  startDrag:        PropTypes.func.isRequired,
```
jstray: This is fine; in the future you could always make this setDrag(true or false) to reduce this to one function.

assets/js/WfParameter.test.js
```diff
@@ -28,5 +29,11 @@ describe('WfParameter', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('Renders string input field', () => {
```
jstray: Hooray for the beginning of WfParameter tests!

assets/js/WfParameter.js
```diff
@@ -69,6 +69,10 @@ export default class WfParameter extends React.Component {
     if (this.type == 'checkbox') {
       this.paramChanged(e.target.checked, DIDNT_PRESS_ENTER)
     }
+
+    if (this.type == 'string' && !this.props.isReadOnly) {
```
jstray: A comment would be helpful here

PR #54, opened 2018-03-23
Conditional UI

harry-tc-zhang:
Functionality for conditions based on menu parameters complete.


assets/js/WfParameter.js
```diff
@@ -277,11 +277,48 @@ export default class WfParameter extends React.Component {
...
+  displayConditionalUI(condition, type) {
+    // Checks if conditional UI should be displayed
+    // type is either 'visible_if' or 'visible_if_not'
+    if(('id_name' in condition) && ('value' in condition)) {
+      var condValues = condition['value'].split('|').map(cond => cond.trim());
+      var selectionIdx = parseInt(this.props.getParamText(condition['id_name']));
+      if(selectionIdx != NaN) {
+        var menuItemsStr = this.props.getParamMenuItems(condition['id_name']);
+        if(menuItemsStr) {
+          var menuItems = menuItemsStr.split('|').map(cond => cond.trim());
```
jstray: Is there any case where you would want the raw string instead of the array from getParamMenuItems? If not, this split could be moved into getParamMenuItems.

assets/js/WfParameter.js
```diff
@@ -277,11 +277,48 @@ export default class WfParameter extends React.Component {
     }
   }
 
+  displayConditionalUI(condition, type) {
+    // Checks if conditional UI should be displayed
```
jstray: Could use a slightly more detailed comment: checks if the menu item referenced by this parameter is set to a value where we are visible

assets/js/WfParameter.test.js
```diff
@@ -35,5 +48,58 @@ describe('WfParameter', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  // Conditional UI tests
```
jstray: Good tests

server/initmodules.py
```diff
@@ -158,6 +158,18 @@ def load_parameter_spec(d, module_version, order):
     pspec.derived_data = d.get('derived-data', False)
     pspec.placeholder = d.get('placeholder', '')
 
+    # For now I will only allow one of them to be active to avoid logically weird stuff.
```
jstray: Yes, they can definitely contradict. I think I would prefer one "visible_if" field and one "invert_visible_condition" boolean. This seems like it would simplify everything downstream, as well as make it impossible for the visible_if and visible_if_not fields to get into a conflicting state.

server/tests/test_initmodules.py
```diff
@@ -222,6 +246,24 @@ def test_module_reload(self):
...
 
+    # A brief check of conditional UI, in that the JSON can be stored and retrieved correctly.
+    def test_condui(self):
+        self.assertEqual(len(Module.objects.all()), 0)
+        load_module_from_dict(self.cond_ui_valid)
+        cond_spec = ParameterSpec.objects.get(id_name='cond_test')
+        cond_spec_visibility = json.loads(cond_spec.visible_if)
+        self.assertEqual(cond_spec_visibility, self.cond_ui_valid['parameters'][1]['visible_if'])
+
+        new_cond_ui = copy.deepcopy(self.cond_ui_valid)
```
jstray: What are you testing with these deep copies? Could use comments

PR #58, opened 2018-03-27
Animate first module on click

mgerring:


mgerring:
modulestack still needs tests, I promise they're coming soon

PR #63, opened 2018-04-09
Conditional UI with booleans, CSS changes for grouped paramaters

mgerring:


assets/js/WfParameter.js
```diff
@@ -40,6 +40,13 @@ export default class WfParameter extends React.Component {
 
     this.type = this.props.p.parameter_spec.type;
     this.name = this.props.p.parameter_spec.name;
+    this.paramClassName = '';
```
jstray: Our hack needs to be commented

assets/css/components/Input_components.scss
```diff
@@ -188,3 +188,38 @@ select.form-control {
...
+.addremove label {
+  @extend .icon;
+  background: transparent;
+  width:2rem;
+  height:2rem;
+  font-size:2rem;
+  display:block;
+}
+
+.addremove .checkbox + label {
```
jstray: I actually don't know what + means in a CSS selector...

PR #65, opened 2018-04-19
Excel formulas

mgerring:
First pass at Excel formulas -- gives us usable formula parsing for formulas that only deal with values in the current row, a la our Python formulas syntax.


server/modules/formula.py
```diff
@@ -14,29 +25,85 @@ def render(wf_module, table):
         if formula == '':
             return table    # nop if no formula
 
-        colnames = [x.replace(" ", "_") for x in table.columns]  # spaces to underscores in column names
-        newcol = pd.Series(np.zeros(len(table)))
+        syntax = wf_module.get_param_menu_idx('syntax')
+
+        newcol = pd.Series(np.zeros(len(table)), dtype=np.dtype(object))
+
+        if syntax == 0:  # Python
```
jstray: Would be good to break Excel and Python cases into separate functions

server/modules/formula.py
```diff
@@ -14,29 +25,85 @@ def render(wf_module, table):
...
+                # Much experimentation went into the form of this loop for good performance.
+                # Note we don't use iterrows or any pandas indexing, and construct the values dict ourselves
+                for i,row in enumerate(table.values):
+                    newcol[i] = eval(code, custom_code_globals, dict(zip(colnames, row)))
+            except Exception as e:
+                return(str(e))
+
+        if syntax == 1:  # Excel
+            try:
+                code = Parser().ast(formula)[1].compile()
```
jstray: what is the [1] here?

server/modules/formula.py
```diff
@@ -14,29 +25,85 @@ def render(wf_module, table):
...
+            except Exception as e:
+                return "Couldn't parse formula: %s" % str(e)
```
jstray: All new strings need to be localizable by passing them to the _ function. See e.g. https://github.com/CJWorkbench/cjworkbench/blob/master/server/dispatch.py#L165

server/tests/modules/test_formula.py
```diff
@@ -1,26 +1,40 @@
...
         self.fpval = ParameterVal.objects.get(parameter_spec=formula_pspec)
+
+        syntax_pspec = ParameterSpec.objects.get(id_name='syntax')
+        self.syntax_pval = ParameterVal.objects.get(parameter_spec=syntax_pspec)
+
         output_pspec = ParameterSpec.objects.get(id_name='out_column')
         self.rpval = ParameterVal.objects.get(parameter_spec=output_pspec)
 
-    def test_formula(self):
+        data_pspec = ParameterSpec.objects.get(id_name='csv')
```
jstray: delete this, you said

server/tests/modules/test_formula.py
```diff
@@ -58,9 +73,76 @@ def test_spaces_to_underscores(self):
...
+        table['formula output'] = table['formula output'].astype(object)
         self.assertTrue(out.equals(table))
 
+    def test_ref_to_number(self):
+        self.assertTrue(letter_ref_to_number('A') == 0)
+        self.assertTrue(letter_ref_to_number('AA') == 26)
+        self.assertTrue(letter_ref_to_number('AZ') == 51)
+        self.assertTrue(letter_ref_to_number('BA') == 52)
+
+    def test_excel_formula(self):
```
jstray: Need to ensure testing all range syntaxes we parse.

PR #68, opened 2018-04-26
Sort from table: all functionalities complete.

harry-tc-zhang:
Hovering over table headers currently does not show anything, due to a limitation with react-data-grid itself. But other stuff work.


assets/js/DataGrid.js
```diff
@@ -152,7 +156,18 @@ export default class DataGrid extends React.Component {
         minWidth={this.state.gridWidth -2}
         minHeight={this.state.gridHeight-2}   // -2 because grid has borders, don't want to expand our parent DOM node
         enableCellSelect={true}
-        onGridRowsUpdated={this.onGridRowsUpdated} />
+        onGridRowsUpdated={this.onGridRowsUpdated}
+        onGridSort={this.onGridSort}
+        // Ref sets the sort direction on the datagrid
+        ref={(component) => {
```
jstray: This code should be broken out into a separate function. Also, not sure this is the right place to put this code... but can't think of a better way.

assets/js/DataGrid.test.js
```diff
@@ -28,12 +28,14 @@ it('Renders the grid', () => {
...
 
   var editCellMock = jest.fn();
+  var sortMock = jest.fn();
 
   const tree = mount( <DataGrid
     totalRows={testData.totalRows}
     columns={testData.columns}
     getRow={getRow}
     onEditCell={editCellMock}
+    onGridSort={sortMock} // I tried but could not get this to work, similar to onEditCell
```
jstray: What was not working? We do need tests

assets/js/EditCells.js
```diff
@@ -49,6 +50,29 @@ function findEditCellsModule(state, wfModuleId) {
...
   return null;
 }
+*/
+
+// A generalized version of the original findEditCellsModule for use in other modules
+export function findModuleWithIdAndIdName(state, wfModuleId, moduleIdName) {
+  var wfModules = state.workflow.wf_modules;
+  var idx = getWfModuleIndexfromId(state, wfModuleId);
+
+  // Is this an existing Edit Cells module?
```
jstray: If you're going to generalize this routine, please update the comments

assets/js/EditCells.js
```diff
@@ -84,7 +108,8 @@ function addEditToEditCellsModule(wfm, edit) {
 export function addCellEdit(wfModuleId, edit) {
   var state = store.getState();
 
-  var existingEditCellsWfm = findEditCellsModule(state, wfModuleId);
+  //var existingEditCellsWfm = findEditCellsModule(state, wfModuleId);
```
jstray: delete this line

assets/js/OutputPane.js
```diff
@@ -158,7 +160,32 @@ export default class OutputPane extends React.Component {
...
   }
 
+  findCurrentModuleInWorkflow(wf) {
+      if(!wf) {
+          return null;
+      }
+      if(!wf.wf_modules) {
+          return null;
+      }
+      var modulesFound = wf.wf_modules.filter((wfm) => {return wfm.id == this.props.selectedWfModuleId});
```
jstray: Can just use find() and check for -1 return value

assets/js/OutputPane.js
```diff
@@ -158,7 +160,32 @@ export default class OutputPane extends React.Component {
       });
   }
 
+  findCurrentModuleInWorkflow(wf) {
```
jstray: This is general purpose and should move into utls

assets/js/OutputPane.js
```diff
@@ -168,6 +195,8 @@ export default class OutputPane extends React.Component {
         resizing={this.state.resizing}
         api={this.props.api}
         setBusySpinner={this.setBusySpinner}
+        sortColumn={moduleIsSort ? currentModule.parameter_vals[0].value : undefined}
```
jstray: Why parameter_vals[0] and [2]? You need to get them by id_name so that this code still works if parameters are re-ordered.

assets/js/OutputPane.js
```diff
@@ -158,7 +160,32 @@ export default class OutputPane extends React.Component {
...
+      if(!wf.wf_modules) {
+          return null;
+      }
+      var modulesFound = wf.wf_modules.filter((wfm) => {return wfm.id == this.props.selectedWfModuleId});
+      return modulesFound.length > 0 ? modulesFound[0] : null;
+  }
+
   render() {
+    // We figure out whether we need to indicate sort status here so that we don't have to
+    // pass a ton of data to the TableView
```
jstray: I'm not super happy putting all this sort code in here, it really belongs in TableView. Can we just pass the currentModule to tableView and let it figure out the sort thing there?

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
...
+import {getPageID} from "./utils";
+import WorkbenchAPI from './WorkbenchAPI'
+import {findModuleWithIdAndIdName, findParamValByIdName, getWfModuleIndexfromId} from './EditCells'
+
+var api = WorkbenchAPI();
+export function mockAPI(mock_api) {
+  api = mock_api;
+}
+
+function getNextSortDirection(current, sortType) {
```
jstray: Too many constants here. Can you define const STRING_TYPE = 0 etc. elsewhere and use throughout? Also I think you could simplify the logic: just increment or decrement dir depending on type, then take mod 3

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react'
+import {store, setSelectedWfModuleAction} from "./workflow-reducer";
+import {getPageID} from "./utils";
+import WorkbenchAPI from './WorkbenchAPI'
+import {findModuleWithIdAndIdName, findParamValByIdName, getWfModuleIndexfromId} from './EditCells'
```
jstray: If these are generally useful they need to be moved out of EditCells and into something else. utils.js? findThings.js?

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
...
+function updateSortModule(wfm, sortInfo, sortType) {
+    var column = sortInfo.column;
+    var columnParam = findParamValByIdName(wfm, "column");
+    var typeParam = findParamValByIdName(wfm, "dtype");
+    // If column changes then we need to change both sort column and type
+    if(columnParam.value != column) {
+        api.onParamChanged(columnParam.id, {value: column})
+            .then(() => {
+                api.onParamChanged(typeParam.id, {value: sortType}).then(() => {
+                    // Timeout to prevent database locked 500 errors.
```
jstray: Remove this timeout. Let's fix this in the back end, either by fixing the locks (changing the timeouts) or by creating an endpoint to update multiple params at once.

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
...
+                        updateSortDirection(wfm, sortInfo, sortType, true);
+                    }, 200);
+                });
+            });
+    } else {
+        updateSortDirection(wfm, sortInfo, sortType);
+    }
+}
+
+export function updateSort(wfModuleId, sortInfo) {
```
jstray: What is inside sortInfo? Is it just the column name? If so, this parameter should just be the column

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
...
+            let sortTypeIdx = sortTypes.indexOf(sortType);
+            var existingSortModule = findModuleWithIdAndIdName(state, wfModuleId, 'sort-from-table')
+            if(existingSortModule) {
+                updateSortModule(existingSortModule, sortInfo, sortTypeIdx);
+                if(existingSortModule.id != wfModuleId) {
+                    store.dispatch(setSelectedWfModuleAction(existingSortModule.id));
+                }
+            } else {
+                let wfModuleIdx = getWfModuleIndexfromId(state, wfModuleId);
+                api.addModule(getPageID(), state.sortModuleId, wfModuleIdx + 1)
```
jstray: This could probably be replaced with an add module action, which also sets the selected module

assets/js/SortFromTable.js
```diff
@@ -0,0 +1,120 @@
...
+                    .then((newWfm) => {
+                        store.dispatch(setSelectedWfModuleAction(newWfm.id))
+                            .then(() => {
+                                // A timeout is set to prevent locking up the database
+                                setTimeout(() => {updateSortModule(newWfm, sortInfo, sortTypeIdx);}, 200);
```
jstray: Again, this bug workaround should not be here.

server/views/WfModule.py
```diff
@@ -247,6 +247,40 @@ def wfmodule_input(request, pk, format=None):
...
+        if not wf_module.workflow.user_authorized_read(request.user):
+            return HttpResponseForbidden()
+
+        table = execute_wfmodule(wf_module)
+        dtypes = table.dtypes.to_dict()
+        ret_types = []
+        for col in dtypes:
+            # We are simplifying the data types here.
+            # More stuff can be added to these lists if we run into anything new.
+            stype = "String"
```
jstray: This logic should probably go in the usual render() results 
- because we already use a zero-row render to get column names for column parameters (see WorkbenchAPI.inputColumns) so we shouldn't duplicate that functionality here
- so other modules can use returned column types without having to make a separate API call

jstray:
Manually merged / conflicts resolved.

PR #76, opened 2018-05-10
Lesson track current step

adamhooper:
https://www.pivotaltracker.com/story/show/157227412

The logic:

* LessonSection tracks this.state.activeStepIndex, default 0
* a Redux reducer sets this.props.stateVersion, a monotonically-increasing number, whenever state changes
* after `render()` _and_ `setTimeout(..., 0)`, we assume the DOM is updated: we test each step in turn until we find that one is incomplete and update `this.state.activeStepIndex`
* when `this.state.activeStepIndex` updates, that triggers another `render()`.

We assume `this.state.activeStepIndex` won't change after the second render, so there's no infinite loop.

My findings:

* We _need_ at least one integration test per lesson. There are oodles of potential bugs, and there are many moving parts we don't test elsewhere (e.g., if we nix a `data-module-name` property because we think it isn't needed, or we change a `<textarea>` to `<input>`)
* Writing `document.querySelector()` in an HTML attribute is frustrating: we use `&quot;` a lot. Perhaps we can avoid that using backticks (JS string interpolation syntax).
* ... putting that all together: it isn't easy coding a lesson.


assets/js/lessons/LessonSection.js
```diff
@@ -30,5 +116,49 @@ LessonSection.propTypes = {
   steps: PropTypes.arrayOf(PropTypes.shape({
     html: PropTypes.string.isRequired,
     highlight: LessonHighlightsType.isRequired,
+    testJs: PropTypes.string.isRequired,
   })).isRequired,
 }
+
+function buildMapStateToProps() {
+  // We want our props to change whenever the state changes -- or at least,
```
jstray: Loving the comments

adamhooper:
I believe Travis failed because of an intermittent error.

PR #84, opened 2018-05-25
Reorder from table

harry-tc-zhang:
Functionality is complete although some style refinements will be needed to make the cells being dragged look nicer.


assets/js/DraggableContainer.js
```diff
@@ -0,0 +1,33 @@
+// Pasted from react-data-grid according to this comment:
```
jstray: You shouldn't need to add this file -- it appears to be installable as part of react-data-grid-addons https://github.com/adazzle/react-data-grid/tree/master/packages/react-data-grid-addons

assets/js/DataGrid.test.js
```diff
@@ -27,58 +29,97 @@ it('Renders the grid', () => {
...
-  expect(tree.find('HeaderCell')).toHaveLength(0);
-
-  expect(tree).toMatchSnapshot();
+  it('Renders the grid', () => {
+
+    var editCellMock = jest.fn();
+    var sortMock = jest.fn();
+
+    const tree = mount(
+        <DragDropContextProvider backend={TestBackend}>
```
jstray: Is wrapping in DropContextProvider needed to get the dragging working? I thought that was a RectDnD thing only, so it wouldn't be compatible with DraggableContainer, but I could be wrong.

assets/js/DraggableHeaderCell.js
```diff
@@ -0,0 +1,100 @@
+// Pasted from react-data-grid according to this comment:
```
jstray: This file should be installed with react-data-grid-addons https://github.com/adazzle/react-data-grid/tree/master/packages/react-data-grid-addons

assets/js/OutputPane.test.js
```diff
@@ -34,7 +36,11 @@ describe('OutputPane', () => {
       render: jsonResponseMock(testData),
     };
 
-    const tree = mount(<OutputPane id={100} revision={1} api={api}/>)
+    const tree = mount(
+        <DragDropContextProvider backend={TestBackend}>
```
jstray: Is DragDropContextProvider applicable to DraggableHeader?

assets/js/OutputPane.test.js
```diff
@@ -50,20 +56,28 @@ describe('OutputPane', () => {
...
   it('Iframe when htmloutput set', () => {
-    const tree = mount(<OutputPane
-      id={undefined}
-      workflow={{id:777,public:true}}
-      selectedWfModuleId={999}
-      revision={1}
-      htmlOutput={true}
-      api={{}}/>)
+    const tree = mount(
+        <DragDropContextProvider backend={TestBackend}>
```
jstray: Is DragDropContextProvider applicable to DraggableHeader?

assets/js/ReorderColumns.test.js
```diff
@@ -0,0 +1,194 @@
...
+    });
+
+    it('Adds a new reorder module after the current non-reorder module '
+        + 'if a reorder module does not exist next to it and sets parameters correctly', (done) => {
+        updateReorder(LOADURL_WFM_ID, {
+            column: 'test_col',
+            from: 3,
+            to: 0
+        });
+        setImmediate(() => {
```
jstray: I don't think you need the setImmediate here, as you're not waiting for a promise to resolve (nothing here depends on the result from a mocked API call)

assets/js/ReorderColumns.test.js
```diff
@@ -0,0 +1,194 @@
...
+        });
+    });
+
+    it('Updates the parameter values of an adjacent reorder module correctly', (done) => {
+        updateReorder(FILTER_WFM_ID, {
+            column: 'test_col',
+            from: 3,
+            to: 0
+        });
+        setImmediate(() => {
```
jstray: Unnecessary setImmediate?

assets/js/ReorderColumns.test.js
```diff
@@ -0,0 +1,194 @@
...
+        });
+    });
+
+    it('Updates the parameter values of the currently selected reorder module correctly', (done) => {
+        updateReorder(REORDER_WFM_ID, {
+            column: 'test_col',
+            from: 3,
+            to: 0
+        });
+        setImmediate(() => {
```
jstray: Unnecessary setImmediate?

assets/js/TableView.test.js
```diff
@@ -24,20 +26,35 @@ describe('TableView', () => {
...
+    let addCellEditMock = jest.fn();
+    mockAddCellEdit(addCellEditMock);
+    let updateSortMock = jest.fn();
+    mockSortColumn(updateSortMock);
+    let reorderColumnsMock = jest.fn();
+    mockReorderColumns(reorderColumnsMock);
+
+
+    const tree = mount(
+        <DragDropContextProvider backend={TestBackend}>
```
jstray: We've recently upgraded to Enzyme 3. Now `shallow()` will call lifecycle method so we shouldn't need `mount()`. Which means the `DataGrid` child will not be mounted, so we shouldn't need `DragDropContextProvider`

assets/js/test-utils.js
```diff
@@ -27,6 +29,16 @@ export function okResponseMock () {
   return jsonResponseMock(null)
 }
 
+export function wrapInTestContext(DecoratedComponent) {
```
jstray: I don't see this used anywhere, am I missing something?

server/modules/reorder.py
```diff
@@ -0,0 +1,36 @@
+from .moduleimpl import ModuleImpl
+import json
+
+
+class ReorderFromTable(ModuleImpl):
+    @staticmethod
+    def render(wf_module, table):
+        # Error message for when reorder history is corrupted
+        CORRUPT_HISTORY_ERROR = 'Reorder history is corrupted. Did you remove one of the input columns?'
```
jstray: Does it not handle removal of input columns? It needs to, this will happen all the time.

server/tests/modules/test_reorder.py
```diff
@@ -0,0 +1,95 @@
...
+        self.test_csv = (
+            'name,date,count,float\n'
+            + 'Dolores,2018-04-22,3,13.5\n'
+            + 'Bernard,2018-04-23,not_a_count,2.0\n'
+            + 'Ford,2016-10-02,5,\n'
+            + 'Dolores,0_not_a_date,4,2.8'
+        )
+        # A reference table for correctness checking
+        self.table = pd.read_csv(io.StringIO(self.test_csv), dtype={
+            'name': object,
```
jstray: Types will probably come through correctly in this test without explicitly setting them here. Try it!

PR #85, opened 2018-06-08
Rename columns

harry-tc-zhang:
At some point travis build started failing somewhere after the npm and python tests and I'm not sure why. Otherwise, everything is working at least on my end.


assets/js/wfparameters/RenameEntries.js
```diff
@@ -0,0 +1,194 @@
...
+
+        this.state = {
+            inputValue: this.props.newColname
+        };
+
+        this.handleChange = this.handleChange.bind(this);
+        this.handleKeyPress = this.handleKeyPress.bind(this);
+        this.handleBlur = this.handleBlur.bind(this);
+        this.handleFocus = this.handleFocus.bind(this);
+        this.handleDelete = this.handleDelete.bind(this);
```
adamhooper: Babel gives us a fancier way of doing this. Instead of a whole bunch of `.bind` methods in the ctor, just declare bound methods this way:

```javascript
handleDelete = (ev) => {
  ...
}
```

No need to change this code. But I think the fancy way should be our standard in the future.

assets/js/wfparameters/RenameEntries.js
```diff
@@ -0,0 +1,194 @@
...
+    handleDelete() {
+        this.props.onEntryDelete(this.props.colname);
+    }
+
+    render() {
+        // The class names below are used in testing.
+        // Changing them would require updating the tests accordingly.
+        return (
+            <div>
+                <div className={'rename-column'} style={{width: '40%', float: 'left'}}>{this.props.colname}</div>
```
adamhooper: Do not use `style`. Use CSS.

Passing object parameters like this forces DOM writes with every render, even when nothing changes.

assets/js/wfparameters/RenameEntries.js
```diff
@@ -0,0 +1,194 @@
...
+                <input
+                    className={'rename-input'}
+                    style={{width: '50%'}}
```
adamhooper: Ditto: nix `style`

package.json
```diff
@@ -36,7 +36,8 @@
     "redux-promise-middleware": "^5.1.1",
     "redux-thunk": "^2.3.0",
     "reselect": "^3.0.1",
-    "sizeof": "^1.0.0"
+    "sizeof": "^1.0.0",
+    "superagent": "^3.8.3"
```
adamhooper: This looks like a bad merge. (I nixed `superagent` Friday.) Did you add code that requires `superagent`? If not, please `npm remove` it.

assets/js/RenameColumns.test.js
```diff
@@ -0,0 +1,240 @@
...
+
+            // Checks the parameter setting part, which should have two parts:
+            // 1. Setting loadAll to false
+            // 2. Setting the new rename entries
+            expect(api.onParamChanged.mock.calls).toHaveLength(2);
+            // Check that we set loadAll to false
+            expect(api.onParamChanged.mock.calls[0][0]).toBe(NEW_RENAME_DISPLAY_ALL_PAR_ID);
+            expect(api.onParamChanged.mock.calls[0][1].value).toBe(false);
+            // Check that we set the new entries
+            expect(api.onParamChanged.mock.calls[1][0]).toBe(NEW_RENAME_ENTRIES_ID);
```
adamhooper: These two lines can replace _all_ ther other `expect` calls on `api.onParamChanged`.
```javascript
expect(api.onParamChanged).toHaveBeenNthCalledWith(1, NEW_RENAME_DISPLAY_ALL_PAR_ID, false);
expect(api.onParamChanged).toHaveBeenNthCalledWith(2, NEW_RENAME_ENTRIES_ID, [ { cornerstone: 'cs' } ]);
```

Ditto above: `expect(api.addModule).toHaveBeenCalledWith([fill me in], initialState.renameModuleId, 1)`

assets/js/WfParameter.js
```diff
@@ -272,6 +273,17 @@ export default class WfParameter extends React.Component {
           history={this.props.getParamText('reorder-history')}
         />
       )
+    } else if (id_name == 'rename-entries') {
+      return (
+          <RenameEntries
+              loadAll={this.props.getParamText('display-all')}
+              changeLoadAll={(val) => {this.props.setParamText('display-all', val)}}
```
adamhooper: Make this a method on `WfParameter`. Otherwise, every re-render of `WfParameter` causes a re-render of `<RenameEntries>`, even when nothing has changed.

assets/js/DataGrid.js
```diff
@@ -94,6 +94,90 @@ class ReorderColumnDropZone extends React.PureComponent {
...
+  }
+
+  handleInputBlur() {
+    this.handleInputCommit();
+  };
+
+  handleInputKeyPress(event) {
+    if(event.key == 'Enter') {
+      this.handleInputCommit();
+    }
```
adamhooper: I think we should handle 'Escape' here: just do `setState()` back to the input props.

assets/js/wfparameters/RenameEntries.test.js
```diff
@@ -0,0 +1,194 @@
...
+    it('Updates parameter upon input completion via blur', (done) => {
+        let tree = mount(<RenameEntries
+            loadAll={false}
+            changeLoadAll={jest.fn()}
+            entries={JSON.stringify(testEntries)}
+            wfModuleId={WFM_ID}
+            paramId={PARAM_ID}
+        />);
+
+        setImmediate(() => {
```
adamhooper: Next time, I think we should use async-function tests instead of async-callback tests, since async tests add unnecessary nesting. See `GoogleFileSelect.test.js` for an example.

assets/js/RenameColumns.js
```diff
@@ -0,0 +1,67 @@
...
+            api.onParamChanged(showAllParam.id, {value: false})
+                .then(api.onParamChanged(entriesParam.id, {value: JSON.stringify(existingEntries)}));
+        } catch(e) {}
+    }
+    else {
+        api.onParamChanged(entriesParam.id, {value: JSON.stringify(existingEntries)});
+    }
+}
+
+export function updateRename(wfModuleId, renameInfo) {
```
adamhooper: This should clearly be a reducer. But I see that we use this exact pattern in all our other hard-coded DataGrid-related modules, so I guess we can defer on fixing that.

assets/js/DataGrid.js
```diff
@@ -513,6 +534,10 @@ export default class DataGrid extends React.Component {
       console.log('More than one row changed at a time in DataGrid, how?')
     }
 
+    if(this.props.isReadOnly) {
+      return;
```
adamhooper: We should never get here, right?

If that's the case, change `return` to `throw new Error(...)`

assets/js/DataGrid.js
```diff
@@ -111,12 +112,14 @@ export class EditableColumnName extends React.Component {
...
     this.handleInputChange = this.handleInputChange.bind(this);
     this.handleInputBlur = this.handleInputBlur.bind(this);
-    this.handleInputKeyPress = this.handleInputKeyPress.bind(this);
+    this.handleInputKeyDown = this.handleInputKeyDown.bind(this);
     this.handleInputFocus = this.handleInputFocus.bind(this);
   }
 
-  enterEditMode() {
-    this.setState({editMode: true});
+  enterEditMode(mode=true) {
```
adamhooper: Do not add a `mode` parameter here. `enterEditMode(false)` does exactly the opposite of what I expect.

Two alternatives: a new `exitEditMode()` method, or changing it to `setIsEditing(isEditing)` with no default. Either way, you're stuck with two methods. No single method name can intuitively represent the two verbs, "start editing in response to a click" and "stop editing because I am the programmer and I say so."

assets/js/WfParameter.js
```diff
@@ -274,14 +274,16 @@ export default class WfParameter extends React.Component {
         />
       )
     } else if (id_name == 'rename-entries') {
+      const changeLoadAll = ((val) => {this.props.setParamText('display-all', val)});
```
adamhooper: Move this up so that it's a property of the `WfParameter`, the way `paramChanged` is. Give it as verbose a name as it needs -- `setParamText_DisplayAll()`, for instance.

Here's why: where it is now, each time we reach this line we create a new `changeLoadAll` function, which means we change that prop on `RenameEntries`. The prop change forces a `render()`.

There are many ways to avoid a render. By far the simplest and most widely-used is to move `changeLoadAll` to the class level. (Yes, it makes `WfParameter.js` more terrible. But that's because `WfParameter` is inescapably terrible.) Then we can make `RenameEntries` extend `React.PureComponent` and presto! No more re-renders.

PR #88, opened 2018-07-20
Refactored functions to update modules from the table

emrig:
Refactored functions to update modules from the table into UpdateTableAction.js.

Removed:
-SortFromTable.js
-ReorderColumns.js
-RenameColumns.js
-DuplicateFromTable.js

Unit tests still broken out into individual js files per function.

Function and moduleId mapping in UpdateTableAction.js is a little funky (`updateModuleMapping`), but didn't want to update any server-side code as my PyCharm search all was not working correctly (Worried about breaking dependencies). Let me know if you have better ideas.


emrig:
Opening new request with additional refactoring

PR #90, opened 2018-07-20
From table refactor

emrig:
Removed individual module_id caching functions in workflows.py into one function driven by dynamic module mappings. Old Id alias' retained in addition to new objects because there are several dependencies.


assets/js/DataGrid.test.js
```diff
@@ -28,6 +28,9 @@ describe('DataGrid tests,', () => {
     ]
   };
 
+  var sortDirectionMock = jest.fn()
```
jstray: This is OK, as long as no test checks to see if the function mock was called (the same mock will be used in all places). If that becomes a problem we can put the initialization inside `beforeEach`

server/views/workflows.py
```diff
@@ -99,11 +54,12 @@ def make_init_state(request, workflow=None, modules=None):
...
     if workflow and not workflow.request_read_only(request):
-        ret['editCellsModuleId'] = edit_cells_module_id()
-        ret['sortModuleId'] = sort_module_id()
-        ret['reorderModuleId'] = reorder_module_id()
-        ret['renameModuleId'] = rename_module_id()
-        ret['duplicateModuleId'] = duplicate_module_id()
+        ret['updateTableModuleIds'] = {}
+        for id_name, id_alias in module_id_mapping.items():
+            # Numerous dependencies, keeping ModuleId keys
+            ret[id_alias] = module_id(id_name)
```
jstray: Is this line still needed? It looks like you are writing the module ID to the output twice

PR #95, opened 2018-11-06
Multicolumn update

emrig:
First pass. Render `colnames` instead of `colselect` and update Select Columns modules to match layout placement.

### DIFF
--- a/assets/js/WfParameter.js
+++ b/assets/js/WfParameter.js
@@ -365,7 +365,7 @@ export default class WfParameter extends React.PureComponent {
   render() {
     const { id_name, name, type, visible_if, visible_if_not } = this.props.p.parameter_spec
 
-    // TODO: force display of 'colnames' for now since it will completely replace 'colselect' eventually
+    // TODO: delete the 'colnames' check. Force display of 'colnames' for now since it will completely replace 'colselect' eventually
     if (!this.props.p.visible && id_name !== 'colnames') {
       return null // nothing to see here
     }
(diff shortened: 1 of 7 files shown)


assets/js/WfParameter.js
```diff
@@ -365,7 +365,8 @@ export default class WfParameter extends React.PureComponent {
   render() {
     const { id_name, name, type, visible_if, visible_if_not } = this.props.p.parameter_spec
 
-    if (!this.props.p.visible) {
+    // TODO: force display of 'colnames' for now since it will completely replace 'colselect' eventually
```
adamhooper: This comment says the opposite of what we want. Replace `TODO` with `TODO: delete the 'colnames' check` and then keep the rest of the comment.

assets/js/WfParameter.js
```diff
@@ -500,7 +520,8 @@ export default class WfParameter extends React.PureComponent {
           </div>
         )
 
-      case 'multicolumn':
+      //TODO: Set all multi-column select modules to have type 'multicolumn' for 'colnames'
+      case 'multicolumn' && id_name === 'colnames':
```
adamhooper: I'm surprised this compiles -- I've never heard of this syntax.

Could you please nix the `id_name` test (and with it, that nifty syntax)? As I understand it, we'll never reach this code today (because `multicolumn` render is auto-aborted before we get here), but we want this code in the future. Please correct me if I'm wrong....

So the comment should be: `TODO: [what you wrote]; [2018-11-06] right now this code is never reached, but it will be when we finish cleaning up the `multicolumn` parameter type.

assets/js/wfparameters/ColumnSelector.js
```diff
@@ -7,13 +7,17 @@ export default class ColumnSelector extends React.PureComponent {
   static propTypes = {
     name: PropTypes.string.isRequired,
     isReadOnly: PropTypes.bool.isRequired,
+    initialValue: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
```
adamhooper: With any `func`, please add the signature in a comment. In this case: ` // onChange('A,B') => undefined`

assets/js/wfparameters/ColumnSelector.js
```diff
@@ -82,22 +90,25 @@ export default class ColumnSelector extends React.PureComponent {
...
-          options={columnOptions}
-          menuPortalTarget={document.body}
-          className='react-select'
-          classNamePrefix='react-select'
-          onChange={this.onSelectColumn}
-          value={selectedColumns}
-          isClearable={false}
-          placeholder='Select columns'
-        />
+        <div className="column-multi-select">
```
adamhooper: `multi-select-input-group` would be a better class name. ("Input group" is Bootstrap's name for inputs and buttons that render together.)

assets/js/wfparameters/ColumnSelector.js
```diff
@@ -7,13 +7,17 @@ export default class ColumnSelector extends React.PureComponent {
...
+    onSubmit: PropTypes.func.isRequired,
     value: PropTypes.string.isRequired, // e.g., 'A,B'; may be '' but not null
     allColumns: PropTypes.arrayOf(PropTypes.shape({
       name: PropTypes.string.isRequired
-    })),
-    onChange: PropTypes.func.isRequired // onChange('A,B') => undefined
+    }))
   }
 
+  buttonRef = React.createRef()
```
adamhooper: This seems unused. Nix?

assets/js/wfparameters/ColumnSelector.js
```diff
@@ -62,7 +66,11 @@ export default class ColumnSelector extends React.PureComponent {
...
     ))
-
+    const maybeButton = initialValue === value ? null : (
+      <button title="submit" ref={this.buttonRef} onClick={this.props.onSubmit}>
+        <i className="icon-play" />
+      </button>
+    )
     return (
       // The name attributes in the buttons are used for selection in tests. Do not change them.
       <div>
```
adamhooper: We should put a `className="column-selector"` here (and that's why we shouldn't put `column-multi-select` later... read on....)

(Why not "column-multi-select"? Consistency. Stick with the same name in CSS and JS. If it irks you enough, please pull-request a rename of the entire component :) -- but not in _this_ pull request)

server/modules/selectcolumns.json
```diff
@@ -11,12 +11,19 @@
       "id_name": "colselect",
       "type": "multicolumn",
       "ui-only": true,
+      "visible": false,
```
adamhooper: Oh sonofabitch. This is going to be hellish....

PR #96, opened 2018-11-08
Don't _always_ render after fetch

adamhooper:
This should help with slow cron jobs that nobody checks. Now, we only
render after fetch if we might send an email or if a user has that
workflow open.

### DIFF
--- a/server/models/commands/ChangeDataVersionCommand.py
+++ b/server/models/commands/ChangeDataVersionCommand.py
@@ -1,11 +1,21 @@
+from channels.db import database_sync_to_async
 from django.db import models
 from server.models import Delta, WfModule
+from server import websockets
 from .util import ChangesWfModuleOutputs
 
 
+@database_sync_to_async
+def _workflow_has_notifications(workflow):
+    """Detect whether a workflow sends email on changes."""
+    return workflow.wf_modules.filter(notifications=True).exists()
+
+
 class ChangeDataVersionCommand(Delta, ChangesWfModuleOutputs):
-    wf_module = models.ForeignKey(WfModule, null=True, default=None, blank=True, on_delete=models.SET_DEFAULT)
-    old_version = models.DateTimeField('old_version', null=True)    # may not have had a previous version
+    wf_module = models.ForeignKey(WfModule, null=True, default=None,
+                                  blank=True, on_delete=models.SET_DEFAULT)
+    # may not have had a previous version
+    old_version = models.DateTimeField('old_version', null=True)
     new_version = models.DateTimeField('new_version')
     dependent_wf_module_last_delta_ids = \
         ChangesWfModuleOutputs.dependent_wf_module_last_delta_ids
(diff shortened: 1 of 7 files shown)


server/models/commands/ChangeDataVersionCommand.py
```diff
@@ -20,17 +30,66 @@ def backward_impl(self):
...
+        if await _workflow_has_notifications(self.workflow):
+            await super().schedule_execute()
+        else:
+            await websockets.queue_render_if_listening(
+                self.workflow.id,
+                self.workflow.last_delta_id
+            )
+
+    @classmethod
+    def amend_create_kwargs(cls, *, wf_module, **kwargs):
```
jstray: What is the purpose of putting this here rather than passing it in from `create()` as before?

server/models/commands/ChangeDataVersionCommand.py
```diff
@@ -20,17 +30,66 @@ def backward_impl(self):
...
+        Of course, it's impossible for us to know whether anybody is viewing
+        self.workflow. So we _broadcast_ to them and ask _them_ to request a
+        render if they're listening. This gives N render requests (one per
+        Websockets communicator) instead of 1, but we don't mind because
+        spurious render requests are no-ops.
+
+        From the user's point of view:
+
+            * If I'm viewing self.workflow, changing data versions causes a
+              render. (There isn't even any HTTP traffic: the communicator
```
jstray: What is "the communicator"?

server/views/workflows.py
```diff
@@ -194,7 +194,8 @@ def render_workflow(request: HttpRequest, workflow: Workflow):
             #
```
jstray: Comment above says "edge cases or maybe bugs" but this code path is normal operation for cron, yes?

server/models/commands/ChangeDataVersionCommand.py
adamhooper: Deltas are a minefield of races because Websockets is async and database operations are sync. (There's no way to hold a database lock from an async method.) So Delta application kinda goes like this:

1. Call `create()` (async method -- database access not allowed)
2. `create()` calls `create_impl()` (mostly for historical race-culling reasons; for now the one remaining feature is that commands can give `create()` positional arguments, and `create_impl()` uses kwargs like a normal Django model.)
3. `create_impl()` _locks the workflow_ (we're going sync now) and:
    a. Calls `amend_create_args()` to gather pre-Delta state -- this is the only safe spot to query the database
    b. Calls the _actual_ `.objects.create()`
    c. Calls `forward_impl()` (the command-specific instructions to commit the Delta)
    d. Sets `workflow.last_delta`
    e. Calls `.load_ws_data()`, which queries the database to gather the data we'll be sending over WebSockets
4. Notify WebSockets (async method -- database access not allowed)
5. Schedule a render over RabbitMQ (async method -- database access not allowed)

Long story short: querying the database in `create()` connects to the database from an executor thread and never releases the connection. It's a leak. (The max number of leaked connections is the size of the executor thread pool). Come to think of it, we might have a fair number of such bugs.

There's already a fair amount of documentation to this effect in `Delta`; and in the `tabs` branch things are even cleaner.

server/models/commands/ChangeDataVersionCommand.py
adamhooper: Django-channels' Websockets consumer

https://channels.readthedocs.io/en/latest/topics/testing.html#websocketcommunicator

Oooh, actually it looks like "communicator" is the name of the test interface. I'll change that to "consumer".

server/views/workflows.py
adamhooper: Ooh good catch. I'll add that to the comment.

PR #99, opened 2018-12-13
Add Hext as Workbench pipenv dependency

brandonrobertz:
This PR adds Hext version 0.2.0 as a Workbench dependency. This is all that's required to get the [extractor module](https://github.com/brandonrobertz/autoscrape-extractor-workbench) working.


adamhooper:
(Leaving a comment for future reference: the reason we're doing this is to support @brandonrobertz's upcoming "Hext Scraper"-type module, still in production.

PR #103, opened 2019-02-22
Multi sort

emrig:


server/modules/sort.py
```diff
@@ -32,11 +44,46 @@ def _do_render(table, column, is_ascending):
...
+        sort_columns: [
+            { colname: 'A', is_ascending: True },
+            { colname: 'B', is_ascending: False}
+        ],
+        keep_top: '2'
+    }
+
+    """
+    if 'sort_columns' not in params:
+        try:
```
adamhooper: Please put this block in a separate function, `_migrate_params_v0_to_v1()`. Otherwise we'll need to move the code _later_ (when we add v2, v3, v4....) and that goes against the spirit of write-once, leave-forever.

PR #108, opened 2019-03-05
per-capita: wire up `data-test` attributes

adamhooper:
I also suggested a bit of extra copy in HTML comments. Got carried away, sorry ;)


adamhooper:
@pierreconti I sorta got carried away when writing the tests, because so many different formulas can mean the same thing. I added a bit of copy to suggest alternate formulas in `<!-- HTML comments -->`. It might help with comprehension -- especially to mention "percentage change = percentage - 100"

I also nixed a sentence as we mentioned on #design.

adamhooper:
@pierreconti Also: you could consider the phrasing in "Ratio" to be: `Relative to its previous size, which county has the largest population?` ... I prefer it because it doesn't suggest that we're looking for "growth" (which we cover in "Percent change").

PR #109, opened 2019-03-05
164405747 course move all banner images

adamhooper:
I've moved every lesson-specific image out of `assets/images/` and into ``server/courses/*/*/*.{png,svg}`


adamhooper:
The one exception: `long-and-wide-format/banner.svg` has the wrong URL, and @jstray is fixing that right now :)

PR #115, opened 2019-04-07
Ignore node deps when using develop-module

brandonrobertz:
This prevents a hang/very slow reload times with developing a module using Node.js/npm.


adamhooper:
LGTM. At this point, I'm beginning to think `.gitignore` would be better logic. But let's cross that bridge _next_ time this issue crops up.

PR #116, opened 2019-04-09
Arrayparams

jstray:
First pass on "List" parameter type. No styling yet. Use like this in the module YAML parameters section:

```
    - name: "Repeat This"
      id_name: listme
      type: list
      parameters:
        - id_name: pA
          name: "Luke Vaxhacker"
          type: checkbox
          default: False
        - id_name: pB
          name: "Darth Scheduler"
          type: checkbox
          default: true
        - id_name: pc
          name: "R2D2.5"
          type: string
          default: "looking for my C<3P0"
```


server/models/module_spec_schema.yaml
```diff
@@ -47,12 +47,12 @@ properties:
...
         description: "Specification for how to give a form to the user and give values to the fetch/render functions"
         items:
-            $ref: '#/definitions/ftypes/any'
+            $ref: '#/definitions/param_types/any'
 
 definitions:
-    # FTypes: Form-element types. Each of these is a React component.
-    ftypes:
-        # _lib: properties that are used in more than one ftype.
+    # param_types: Types that module author can specify, which in turn define both the UI and the storage format (DType)
```
adamhooper: `param_types` is not an improvement. The confusing part of "FType" isn't the "F": it's the word, "type".

Because it's not a type. The thing we're describing here is the entire contents of the "parameters" field. One element of that description is the "type" property; but we want to communicate that each definition is for _all the properties together_.

This should be `param_field` ... or maybe `field`? `parameter`?

server/models/module_spec_schema.yaml
```diff
@@ -226,99 +226,121 @@ definitions:
...
             properties:
-                <<: *common_ftype_properties
+                <<: *common_param_type_properties
                 type: { const: custom }
-                name: { $ref: '#/definitions/ftypes/_lib/name' }
-                placeholder: { $ref: '#/definitions/ftypes/_lib/placeholder' }
+                name: { $ref: '#/definitions/param_types/_lib/name' }
+                placeholder: { $ref: '#/definitions/param_types/_lib/placeholder' }
+
+        anyExceptList:
```
adamhooper: `anyExceptList` irks me. While renaming `param_types`, can we ponder this for a second? `anyExceptList` describes the _value_, but not its _purpose_ -- and both will change.

Soon enough, we're going to have _groups_ of parameters. Lists can contain groups but not lists; Groups can't contain groups or lists. At that point, `anyExceptList` will be ambiguous. 

The name is for things that go in lists, right? How about `anyInList`? `anyListChild`? (I love `anyListChild`. Goes nicely with `anyGroupChild`.)

server/models/param_field.py
```diff
@@ -563,6 +573,7 @@ def from_dict(self, d: Dict[str, Any]) -> 'ParamField':
             placeholder=d.get('placeholder', ''),
             tab_parameter=d.get('tab_parameter', ''),
             visible_if=d.get('visible_if'),
+            child_parameters=d.get('parameters', []), # just `parameters` in YAML spec but child_parameters everywhere else
```
adamhooper: You've told me you prefer `parameters` in the YAML spec to `child_parameters`. Is `parameters` better? If so, why? Otherwise, please rename to `child_parameters` here and in the schema.

Also, nit: nix the `, []` because we want `None` on non-List param fields.

server/models/module_spec_schema.yaml
jstray: All I can say is I found the "f" confusing. Working through all of this, I maintain that the "ftype" *is* the parameter type. It's what's specified in the YAML, it completely defines the dtype (and not the other way around), and it's what the user sees. So I think just calling it the "type" is both simple and accurate.

adamhooper: I'm not debating which name is best. I'm requesting a more-consistent pull request.

Prior to this pull request, there are three names for the same Thing:

* module spec: "parameter"
* module schema: "ftype"
* python: "ParamField"
* javascript: "parameter"

Once we merge this pull request (as it is above), there will _still_ be three names for the same Thing:

* module spec: "parameter"
* module schema: "param_type"
* python: "ParamField"
* javascript: "parameter"

You correctly identified a problem in my original code: I named the same Thing three different names. If we're going to go through the trouble of renaming, let's fix that problem.

PR #125, opened 2019-05-19
[courses] Fix continue typo

hydrosquall:
Just learned about Workbench this weekend from a colleague, it's a great idea! Here's a small text fix for you.


adamhooper:
Thanks!

PR #135, opened 2019-08-30
Add a pluralized message to catalogs

giorgio93p:


adamhooper:
This looks like a pattern we can maintain for a long while. Thank you for figuring out all the little details.

PR #137, opened 2019-09-25
Add django template tag for custom i18n

giorgio93p:


cjworkbench/i18n/trans.py
```diff
@@ -0,0 +1,181 @@
...
+        except ValueError as error:
+            raise BadCatalogsError(
+                "The catalog for the given locale (%s) is badly formatted" % locale
+            ) from error
+        except Exception as error:
+            raise UnsupportedLocaleError(
+                "Can't load a catalog for the given locale (%s)" % locale
+            ) from error
+
+    def trans(self, message_id, default=None, context=None, parameters={}, tags={}):
```
adamhooper: Can we call it `def trans_html(...)`? The output is valid HTML, and that's a key feature. It should be part of the function signature.

cjworkbench/i18n/trans.py
```diff
@@ -0,0 +1,181 @@
...
+        HTML-like tags in the message used are replaced by their counterpart in `tags`, as specified in `restore_tags`
+        """
+        return self.process_message(
+            self.get_message(message_id, context=context),
+            default or message_id,
+            parameters,
+            tags,
+        )
+
+    def process_message(self, message, plain, parameters={}, tags={}):
```
adamhooper: Should we call it `fallback`, not `plain`? I was confused when reading it.

cjworkbench/tests/i18n.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_format_invalid_message(self):
+        """Tests that a translator can't break our system by including a numeric variable in the message
+        """
+        self.assertEqual(
+            MessageTranslator(default_locale).process_message(
+                "Hello {a} {0} {b}", "Hello {a} {b}", parameters={"a": "you", "b": "!"}
+            ),
+            "Hello you !",
+        )
```
adamhooper: We need more unit tests:

```
Disallow nesting:
process_message("Allo <a0><a1>x</a1></a0>", "Hello <a0>x</a0><a1>y</a1>", tags={"a0": {"tag": "a": "href": "http://a.com"}, "a1": {"tag": "a", "href": "http://b.com"}})
'Allo <a href="http://a.com">&lt;a1&gt;x&lt;/a1&gt;</a>'
```

```
Substitute parameters within tags:
process_message("Allo <a0>x{y}z</a0>", "Hello <a0>x</a0>{y}", tags={"a0": {"tag": "a", "href": "http://a.com"}}, arguments={"y": "Y"})
'Allo <a href="http://a.com">xYz</a>'
```

```
HTML-escape special characters in message:
process_message("Allo < there", "Hello there")
"Allo &lt; there"
```

```
Avoid injection by translators (really, an explicit _reason_ for escaping HTML from translators):
process_message("Allo <script>x=y</script> there", "Hello there")
"Allo &lt;script&gt;x=y&lt;/script&gt; there"
```

```
HTML-escape parameters:
process_message("Allo {script}", "Hello {script}", parameters={"script": "<script>x=y</script>"})
"Allo &lt;script&gt;x=y&lt;/script&gt;"
```

```
HTML-escape arguments
process_message("Allo <a0>x</a0>", "Hello <a0>x</a0>", tags={"a0": {"tag": "a", "href": "/a?x=y&z=b"}})
'Allo <a href="/a?x=y&amp;z=b">x</a>'
```

... your BeautifulSoup algorithm doesn't return these. What _does_ it return? I feel my proposed return values are sensible and easy to explain to translators: "just write plain text; we'll treat &lt;a0&gt; specially."

cjworkbench/i18n/trans.py
```diff
@@ -0,0 +1,181 @@
...
+                return self.format_message(
+                    self.replace_tags(message, tags), parameters=parameters
+                )
+            except Exception:
+                pass
+        return self.format_message(
+            self.replace_tags(plain, tags), parameters=parameters
+        )
+
+    def replace_tags(self, target_message, tags):
```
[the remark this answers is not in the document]
giorgio93p: Indeed. I also renamed `process_message` to `_process_message`, since it's also meant to be used internally.

templates/account/signup.html
```diff
@@ -22,8 +22,9 @@
       </div>
       {% include "socialaccount/snippets/login_extra.html" %}
     {% else %}
-      <p>{% blocktrans %}If you have not created an account yet, then please
-      <a href="{{ signup_url }}">sign up</a> first.{% endblocktrans %}</p>
+      <p>
+        {% trans "accounts.sign_up.login_prompt" default='If you have already created an account, then please <a0>log in</a0>.' tag_a0_href=login_url %}
```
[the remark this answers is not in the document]
giorgio93p: By the way, I noticed that `signup_url` was not defined but `login_url` was. Since prompting people to go the page they are seemed unnecessary, I assumed it was a mistake and changed the whole phrase.

cjworkbench/tests/i18n.py
giorgio93p: Most of these properties ('HTML-escape special characters in message', 'Disallow nesting' but see note 1 below, 'Avoid injection by translators') are already tested in the tests for the `trans_html` tag.

1. For nesting, it's easier to just ignore the inner tags (now we keep their contents, but we could not keep them); for example `<a0><span0>Hello</span0></a0>` would become `<a href="something">Hello</a>` (or plain `<a href-"something"></a>`).
2. Arguments and parameters are provided by the programmer, so it's different decision whether to escape them or not.

adamhooper: Arguments and parameters are variables. They can be provided by the user. (For instance: a column name might be the argument in a message, and column names are provided by users.)

giorgio93p: Nope, arguments and parameters are provided _to `trans`_ by the programmer. If the user inputs them somewhere else, isn't it the duty of that "somewhere else" to decide whether to escape them or not?

adamhooper: No. It is _never_ sensible to _not_ HTML-escape these arguments. If `trans` didn't HTML-escape all arguments, then programmers would need to HTML-escape arguments every. single. time they call it.

This isn't just theoretical. Your own patch is inconsistent. As a programmer, you coded:

```
{% trans "accounts.sign_up.login_prompt" default='If you have already created an account, then please <a0>log in</a0>.' tag_a0_href=login_url %}
```

`login_url` will often contain an ampersand -- which is invalid HTML. It makes no sense to ask programmers to html_escape every single argument. That requirement would give zero benefit, cost callers space and time, and introduce bugs because programmers would never imagine they need to HTML-escape things.

In the example you wrote, the login URL doesn't contain anything the user writes; but that's just happenstance. Parameters can easily contain user-input data. Imagine:

```
process_message("Download this workflow as a <a0>spreadsheet</a0>", ..., tags={"a0": {"tag": "a", "href": csv_url, "download": filesystem_safe_workflow_title, "title": workflow_title}})
```

In this example, `workflow_title` is provided by Bob, and Alice is viewing the page. I'd expect the programmer to _filesystem-escape_ `filesystem_safe_workflow_title` (removing any `/`). But I wouldn't expect them to HTML-escape it afterwards, because that's the whole point of an HTML-producing library. If Bob names the workflow `r&b`, Alice should see the `download` attribute, `download="r&amp;b.csv`.

And if Bob user names the workflow, `"><script>window.x = y</script>`? I'd expect Alice to see a `title` tag of, `"><script>window.x = y</script>`. If `trans` _doesn't_ HTML-escape, then Bob just injected a script in Alice's browser session.1

`trans` should _never_ output invalid HTML. It should behave as programmers are used to:

```javascript
const a = document.createHTMLElement("a")
a.href = csv_url
a.download = filesystem_safe_workflow_title // no need to HTML-escape
a.title = workflow_title // no need to HTML-escape
document.body.appendChild(a)
```

giorgio93p: OK, I get it. I updated my pull request; see if it's fine now.

templates/account/signup.html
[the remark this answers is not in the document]
adamhooper: I think you got this right. These `account`-related templates are a nightmare. We copied them from a dead Django app we still depend on, and then there have been 18 months' worth of changes, mostly untested.

cjworkbench/tests/i18n.py
adamhooper: Still missing a couple:

* "Disallow nesting"
* "HTML-escape special characters in message"

I'm okay with making the retval from "disallow nesting" something a bit different -- so long as it is utterly impossible to nest tags.

HTML-escaping special characters in the message is essential. Otherwise translators could inject XML entities (which would be one-half of a security vulnerability).

giorgio93p: These two are tested in `server/tests/templatetags`

giorgio93p: I moved most tests from `server/tests/templatetags/test_trans.py` to `cjworkbench/tests/i18n.py`. I also added tests for more properties. I think it's fine now.

cjworkbench/tests/i18n.py
```diff
@@ -9,6 +9,27 @@
 
 
 class TransTest(SimpleTestCase):
+    def test_trans_no_default(self):
```
adamhooper: Could you please describe the case where it's okay for a `trans()` call to have no `default`? (I had expected a 500 error and crashed Babel-extract if there's no `default`, so this test expectation comes as a surprise to me.)

giorgio93p: I wanted to include the ability to use IDs themselves as messages, but indeed this is not what we will do in this project. I will change the behavior and the test.

adamhooper:
Woo-hoo! Merged. Thank you for the extraordinary amount of effort you invested in this function.

PR #138, opened 2019-10-07
Translations of django forms

giorgio93p:
This PR inserts the capability of translating the placeholders of django forms. As a side effect, it allows translations in python code and makes sure they extracted correctly. Since django forms need lazy translations, we also need to have this option in our custom i18n code.

### DIFF
--- a/cjworkbench/i18n/extract.py
+++ b/cjworkbench/i18n/extract.py
@@ -200,7 +200,7 @@ def extract_python(fileobj, keywords, comment_tags, options):
                         context = messages[2]
                     else:
                         context = None
-                    
+
                     if context:
                         # if we have a context, trick pybabel to use `pgettext`
                         # so that it adds the context to the translation file
--- a/cjworkbench/i18n/extract.py
+++ b/cjworkbench/i18n/extract.py
@@ -100,8 +100,9 @@ def extract_python(fileobj, keywords, comment_tags, options):
     It returns an iterator yielding tuples in the following form ``(lineno,
     funcname, message, comments)``.
     
-    
-    Adapted from the corresponding pybabel built-in function
+    Adapted from the corresponding pybabel built-in function,
+    so that it understands the syntax of our custom `trans`/`trans_lazy` function
+    and correctly parses the default message and the context.
     
     :param fileobj: the seekable, file-like object the messages should be
                     extracted from
@@ -183,18 +184,26 @@ def extract_python(fileobj, keywords, comment_tags, options):
                     translator_comments = []
 
                 ### HERE start our modifications to pybabel's script
-                if len(messages) > 1 and messages[1]:
-                    translator_comments.append(
-                        (message_lineno, "default-message: " + messages[1])
-                    )
-
-                if len(messages) > 2 and isinstance(messages[2], str):
-                    context = messages[2]
-                else:
-                    context = None
-
                 if funcname in ["trans", "trans_lazy"]:
+                    # `messages` will have all the string parameters to our function
+                    # As we specify in the documentation of `trans`,
+                    # the first will be the message ID, the second will be the default message
+                    # and the (optional) third will be the message context
+                    if len(messages) > 1 and messages[1]:
+                        # If we have a default, add it as a special comment
+                        # that will be processed by our `merge_catalogs` script
+                        translator_comments.append(
+                            (message_lineno, "default-message: " + messages[1])
+                        )
+
+                    if len(messages) > 2 and isinstance(messages[2], str):
+                        context = messages[2]
+                    else:
+                        context = None
+                    
                     if context:
+                        # if we have a context, trick pybabel to use `pgettext`
+                        # so that it adds the context to the translation file
                         funcname = "pgettext"
                         messages = [context, messages[0]]
                     else:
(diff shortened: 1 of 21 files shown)


cjworkbench/i18n/extract.py
```diff
@@ -95,3 +93,159 @@ def extract_django(fileobj, keywords, comment_tags, options):
...
                     yield lineno, None, smart_text(g), comments
+
+
+def extract_python(fileobj, keywords, comment_tags, options):
+    """Extract messages from Python source code.
+    It returns an iterator yielding tuples in the following form ``(lineno,
+    funcname, message, comments)``.
+    
+    
+    Adapted from the corresponding pybabel built-in function
```
adamhooper: In the docstring, could you please list every difference between this function and the pybabel built-in function?

cjworkbench/i18n/extract.py
```diff
@@ -95,3 +93,159 @@ def extract_django(fileobj, keywords, comment_tags, options):
...
+                    messages = messages[0]
+                # Comments don't apply unless they immediately preceed the
+                # message
+                if (
+                    translator_comments
+                    and translator_comments[-1][0] < message_lineno - 1
+                ):
+                    translator_comments = []
+
+                ### HERE start our modifications to pybabel's script
```
adamhooper: (... and what are those modifications?)

cjworkbench/i18n/trans.py
```diff
@@ -42,14 +44,35 @@ def _get_translations(locale):
     return _translators[locale]
 
 
-def trans(locale, message_id, *, default, context=None, parameters={}, tags={}):
+def trans(message_id, *, default, context=None, parameters={}, tags={}):
```
adamhooper: It looks like we're making one function do two things here. Please refactor so we have a total of three functions:

1. `trans()` -- no `context`, no `tags`; outputs normal text. May output unescaped HTML special characters (we escape them elsewhere). Maybe we can even survive without `parameters` support -- meaning, we could maybe wrap Babel's own `trans()` directly. (I'm out of my element here; you may correct me....)
2. `trans_lazy()` -- like `trans()`
3. `trans_html()` -- for our `{% trans %}` template tag -- returns `html_safe()` text. This is the function we worked so hard on -- and it should only be called directly from HTML templates.

My point: `trans_html()` is exclusively for HTML templates; `trans()` is not. They shouldn't be the same function because their users should have different expectations about how they work.

cjworkbench/i18n/trans.py
```diff
@@ -42,14 +44,35 @@ def _get_translations(locale):
...
+    return do_trans(
+        get_language(),
+        default=default,
+        context=context,
+        parameters=parameters,
+        tags=tags,
+    )
+
+
+def do_trans(locale, message_id, *, default, context=None, parameters={}, tags={}):
```
adamhooper: `_do_trans()`

cjworkbench/middleware/i18n.py
```diff
@@ -15,6 +16,7 @@ def __call__(self, request):
             locale = default_locale
 
         request.locale_id = locale
+        activate(locale)
```
adamhooper: Please add a comment explaining why we do this

integrationtests/git-server/modules/columnchart/columnchart.py
```diff
@@ -26,8 +26,8 @@ def python_format_to_d3_tick_format(python_format: str) -> str:
     # Formatter.parse() returns Iterable[(literal, field_name, format_spec,
     # conversion)]
     specifier = next(Formatter().parse(python_format))[2]
-    if not specifier or specifier[-1] not in 'bcdoxXneEfFgGn%':
-        specifier += 'r'
+    if not specifier or specifier[-1] not in "bcdoxXneEfFgGn%":
```
adamhooper: Aaaah -- Black should not reformat these. This was a bug in `master` and I've just fixed it.

Please merge `master` and then undo all edits in `integrationtests/git-server/modules/`. Black won't complain about them again after the update I just pushed to master.

server/templatetags/i18n_icu.py
```diff
@@ -1,6 +1,6 @@
 from django import template
 from django.utils.safestring import mark_safe
-from cjworkbench.i18n.trans import trans as trans_icu
+from cjworkbench.i18n.trans import do_trans
```
adamhooper: `trans_html()`

cjworkbench/i18n/trans.py
giorgio93p: OK, splitting them seems a very good idea.

Babel is written with `gettext` in mind, so we don't want to use it for actually translating stuff.

Regarding `context`: It should also be in `trans()`, since it has to do with message catalogs.

adamhooper: k cool. I was confusing it with Django template-tag context.

giorgio93p:
I have made some changes. Please review them and tell me if anything else needs to be done.

cjworkbench/i18n/trans.py
```diff
@@ -42,14 +44,38 @@ def _get_translations(locale):
     return _translators[locale]
 
 
-def trans(locale, message_id, *, default, context=None, parameters={}, tags={}):
-    """Translate the given message ID to the given locale.
+def trans(message_id, *, default, context=None, parameters={}, tags={}):
```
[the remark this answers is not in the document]
giorgio93p: Done

adamhooper:
Splendid! Thank you :)

PR #143, opened 2019-11-11
Mark a message as fuzzy if its default translation changes

giorgio93p:
The main problem: suppose we change the default translation of a message while keeping the same ID. The translators have to be notified, because the reason for the change may propagate to other languages.
The secondary problem: lingui does not update the default translations of messages unless `--overwrite` is passed as an argument to its extraction script.

The solution to the main problem is the following
- temporarily keep the old catalog of English messages, in the same way we keep the catalogs of the other locales
- find the messages whose translation in English has changed
- mark them (along with the ones that are already marked) as fuzzy in the catalogs of the other locales; I chose to only mark them in the catalogs where they are not empty

Lingui does not provide a solution to the main problem for js messages. There has been some discussion in their issues and the maintainer has said he may add fuzzy messages, but this seems to not even be in the code of the upcoming v3. Hence, we search for fuzzy messages after we have merged the English js and python catalogs.

Pontoon, as well as most of the tools that read and modify `po` files, gives priority to the messages marked as `fuzzy` so that the translators can focus on them first. It also contains a search filter for fuzzy messages.

Since I had to modify the code for catalog merging, I tried to make it more readable in two ways: i) by splitting the script itself better ii) by introducing utility functions, because there were some common operations that required 2-3 lines and cluttered the code.


cjworkbench/i18n/catalogs/util.py
```diff
@@ -0,0 +1,43 @@
...
+def find_corresponding_string(catalog: Catalog, message: Message) -> Optional[str]:
+    """ Search the catalog for a message with the same ID (and context) and return its string.
+    """
+    corresponding_message = find_corresponding_message(catalog, message)
+    return corresponding_message.string if corresponding_message else None
+
+
+def read_po_catalog(filename: str) -> Optional[Catalog]:
+    """ Try to read a po catalog from the given path.
+    Return None on failure
```
adamhooper: Raise an error on failure. None of the callers test for `None` (nor should they). Nix the try/except.

adamhooper:
@giorgio93p Looks like there are conflicts. Could you please merge with master? (Maybe you can look at what I did on master while you're at it.)

PR #144, opened 2019-11-18
Internationalization of (the rest of) js strings

giorgio93p:
I have also merged the branch `fuzzy_messages` here; I believe it will have been merged into `master` until this is ready.


assets/js/util/ModuleCategory.js
```diff
@@ -0,0 +1,16 @@
...
+
+const CategoryNames = {
+  Combine: t('js.util.ModuleCategory.CategoryNames.Combine')`Combine`,
+  Scrape: t('js.util.ModuleCategory.CategoryNames.Scrape')`Scrape`,
+  Clean: t('js.util.ModuleCategory.CategoryNames.Clean')`Clean`,
+  Analyze: t('js.util.ModuleCategory.CategoryNames.Analyze')`Analyze`,
+  Visualize: t('js.util.ModuleCategory.CategoryNames.Visualize')`Visualize`,
+  Code: t('js.util.ModuleCategory.CategoryNames.Code')`Code`,
+  'Add data': t('js.util.ModuleCategory.CategoryNames.AddData')`Add data`
+}
```
giorgio93p: @adamhooper The names of module categories are converted to uppercase when shown as headings in the "Add step" popper, while they are not converted when shown in the success message of importing module from github. Should we ask the translator to define them as uppercase or normal case?

assets/js/params/File/UploadedFileSelect.js
```diff
@@ -108,8 +110,10 @@ const UploadedFileSelect = React.memo(function UploadedFileSelect ({ value, file
 
   return (
     <>
-      <button className='uploaded-file-select' onClick={open} title={canOpen ? 'Choose a previously-uploaded file' : undefined} disabled={!canOpen}>
-        {`File ${valueIndex === -1 ? '[NONE]' : (valueIndex + 1)} of ${files.length}`}
+      <button className='uploaded-file-select' onClick={open} title={canOpen ? i18n._(t('js.params.File.UploadedFileSelect.choosePreviousFile.hoverText')`Choose a previously-uploaded file`) : undefined} disabled={!canOpen}>
+        <Trans id='js.params.File.UploadedFileSelect.choosePreviousFile.text' description='The first parameter is the number of the file (or js.params.File.UploadedFileSelect.choosePreviousFile.none) and the second is the total number of files'>
+            File {valueIndex === -1 ? <Trans id='js.params.File.UploadedFileSelect.choosePreviousFile.none'>[NONE]</Trans> : (valueIndex + 1)} of {files.length}
```
adamhooper: Let's avoid this nested-`<Trans>` pattern. Can we make the translation string look something like `File {fileNumber} of {nFiles}`?

assets/js/params/Multichartseries/index.test.js
```diff
@@ -1,11 +1,12 @@
 /* global describe, it, expect, jest */
 import React from 'react'
 import Multichartseries from './index'
-import { mount } from 'enzyme'
+// import { mount } from 'enzyme'
```
adamhooper: nix

assets/js/params/common/AllNoneButtons.js
```diff
@@ -1,32 +1,41 @@
...
       </button>
     </div>
   )
 }
-AllNoneButtons.propTypes = {
-  isReadOnly: PropTypes.bool.isRequired,
-  onClickNone: PropTypes.func.isRequired, // func() => undefined
-  onClickAll: PropTypes.func.isRequired // func() => undefined
-}
+AllNoneButtons.propTypes =
```
adamhooper: Please move the `{` to this line (and undo your changes to indentation of the following lines).

assets/js/util/ModuleCategory.js
```diff
@@ -0,0 +1,17 @@
...
+  Scrape: t('js.util.ModuleCategory.CategoryNames.Scrape')`Scrape`,
+  Clean: t('js.util.ModuleCategory.CategoryNames.Clean')`Clean`,
+  Analyze: t('js.util.ModuleCategory.CategoryNames.Analyze')`Analyze`,
+  Visualize: t('js.util.ModuleCategory.CategoryNames.Visualize')`Visualize`,
+  Code: t('js.util.ModuleCategory.CategoryNames.Code')`Code`,
+  'Add data': t('js.util.ModuleCategory.CategoryNames.AddData')`Add data`,
+  Other: t('js.util.ModuleCategory.CategoryNames.Other')`Other`
+}
+
+export function getCategoryName (i18n, category) {
```
adamhooper: I find this pattern unpleasant -- not very React-like.

Would this work better as a `withI18n()` React component? Like:

```javascript
const CategoryName = ({ id }) => {
  switch (id) {
    "Combine": return <Trans ...>;
    "Scrape": return <Trans ...>;
    ....
  }
}
```

Would this look nicer from the caller's point of view? (I'm not 100% sure it would be better, but I think it's worth checking.)

assets/js/params/Custom/Code/AceEditor.js
```diff
@@ -74,15 +76,17 @@ export default class WorkbenchAceEditor extends React.PureComponent {
...
-          a <kbd>pd.DataFrame</kbd> and returns
-          a <kbd>pd.DataFrame</kbd>. You may use
-          the <a target='_blank' rel='noopener noreferrer' href='https://docs.python.org/3/library/math.html'><kbd>math</kbd></a>,
-          {' '}<kbd>pd</kbd>{' '}
-          (<a target='_blank' rel='noopener noreferrer' href='https://pandas.pydata.org/pandas-docs/stable/api.html#dataframe'>Pandas</a>)
-          {' '}and <kbd>np</kbd>{' '}
-          (<a target='_blank' rel='noopener noreferrer' href='https://docs.scipy.org/doc/numpy/reference/routines.html'>Numpy</a>)
-          {' '}modules.
+          <Trans id='js.params.Custom.Code.AceEditor.help' description='The parameters contain either code or names of libraries. The tags are either <kbd> formatting or URLs.'>
+              Define a <kbd>{hideFromTrans('process(table)')}</kbd> function that accepts
```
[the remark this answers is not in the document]
giorgio93p: The intention is to make it abundantly clear that this part of the text should not be translated by not making available to the translator. We could write that in a comment, but I think there would be room for misunderstanding and hence a need to check if the translator left these words untranslated, both of which can avoid using `hideFromTrans`.

assets/js/Embed.js
```diff
@@ -74,7 +74,7 @@ export class Embed extends React.Component {
                 <ul>
                   <li>
                     <a href={'/workflows/' + this.props.workflow.id} target='_blank' rel='noopener noreferrer'>
-                    by {this.props.workflow.owner_name}
+                      <Trans id='js.Embed.metadata.author'>by {this.props.workflow.owner_name}</Trans>
```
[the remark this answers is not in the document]
giorgio93p: Indeed, something more descriptive is usually (and in this case) nice to have; however, while replacing, we didn't want to introduce new variables.

assets/js/params/Custom/VersionSelect/DataVersionSelect.js
```diff
@@ -19,27 +25,48 @@ export class DataVersionSelect extends React.PureComponent {
...
+            <Trans id='js.params.Custom.VersionSelect.DataVersionSelect.selectVersion.label'>
+                Version
+            </Trans>
+          </div>
+          <button
+            type='button'
+            title={i18n._(t('js.params.Custom.VersionSelect.DataVersionSelect.selectVersion.hoverText')`Select version`)}
+            onClick={this.handleClickOpenModal}
+          >
+            <Trans id='js.params.Custom.VersionSelect.DataVersionSelect.versionCount' description='The first parameter will be the current version and the second will be the number of versions'>
```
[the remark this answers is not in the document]
giorgio93p: The parameter `{1}` is already named `{nVersions}` in the translation message. I shall modify the comment for clarification.
This string is not actually duplicated, since for some reason in the one case it is all together and in the other case it is split by a button and is in different `<div>`s.

assets/js/params/File/UploadedFileSelect.js
giorgio93p: Since the `{fileNumber}` may be "[NONE]", we can't avoid nesting. We can make it easier for the programmer to read, by splitting it to something like
```
  const nFiles = files.length
  const fileNumber = valueIndex === -1 ? <Trans id='js.params.File.UploadedFileSelect.choosePreviousFile.none'>[NONE]</Trans> : (valueIndex + 1)
        <Trans id='js.params.File.UploadedFileSelect.choosePreviousFile.text' description='The parameter {fileNumber} is the number of the file (or js.params.File.UploadedFileSelect.choosePreviousFile.none) and {nFiles} is the total number of files'>
            File {fileNumber} of {nFiles}
        </Trans>
```

assets/js/WorkflowEditor/DataVersionModal.js
```diff
@@ -5,22 +5,22 @@ import memoize from 'memoize-one'
...
-  'April',
-  'May',
-  'June',
-  'July',
-  'Aug.',
-  'Sept.',
-  'Oct.',
-  'Nov.',
-  'Dec.'
+  /* i18n: AP style month name */t('js.WorkflowEditor.DataVersionModal.Months.Jan')`Jan.`,
```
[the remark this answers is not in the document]
giorgio93p: We could change it now if you like

assets/js/util/ModuleCategory.js
giorgio93p: It would be nicer in some cases, but it's not always feasible: for example, if it were used in an HTML attribute, it would need to be a plain string; also, `ImportModuleForGithub` needs it to be a string (I tried using a react component and got [Object object] in the status message). 

I am adding a React component for use when appropriate, but I will also keep the `getCategoryName` function for the other cases.

assets/js/params/File/UploadApiModal.js
```diff
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types'
 import { Modal, ModalHeader, ModalBody, ModalFooter } from '../../components/Modal'
```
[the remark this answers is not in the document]
giorgio93p: I am not sure, but I believe we might as well translate it for consistency. 

Could a Greek journalist with little knowledge on programming find it more useful if it were translated? I can't tell; maybe we should ask Catherine for input on this?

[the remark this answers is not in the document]
giorgio93p: By the way, how can I see this modal?

assets/js/OutputIframe.js
```diff
@@ -174,7 +175,7 @@ export class OutputIframe extends React.PureComponent {
           <>
             <iframe src={src} />
             <div className='outputpane-iframe-control-overlay'>
-              <button name='embed' className='btn' title='Get an embeddable URL' onClick={this.handleClickOpenEmbedModal}>
+              <button name='embed' className='btn' title={i18n._(t('js.OutputIframe.getEmbeddableUrl.hoverText')`Get an embeddable URL`)} onClick={this.handleClickOpenEmbedModal}>
```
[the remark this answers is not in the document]
giorgio93p: Looks like we haven't finished our discussion from the last time and a few `hoverText`s have already made their way into master.

Both solutions (asking the programmer to either use `hoverText` or add a comment) are hard to enforce. Putting that aside,
- On the one hand, in the long run adding a comment might be a bit easier than renaming the ID. 
- On the other hand, adding an i18n comment requires quite a bit more characters and it must be placed just before `t` (with or without whitespace), so we would need something like
```
<button name='embed' 
  className='btn' 
  title={i18n._(
    /* i18n: This is an HTML title attribute, so it appears on hover */
    t('js.OutputIframe.getEmbeddableUrl.hoverText')`Get an embeddable URL`
  )} 
  onClick={this.handleClickOpenEmbedModal}>
```. Moreover, the `.title` suffix has also been used in the message IDs of some `<h1>`-`<h6>` elements; the translator could think that we forgot to add the hover comment to them, so we would also need to either change those IDs or add comments there as well.

File: assets/js/params/Custom/Code/AceEditor.js (hunk 6 above)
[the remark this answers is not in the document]
Comment (adamhooper) [reply assets/js/params/Custom/Code/AceEditor.js]: In this case, I think translators need to see the contents. It's `a [process()] function` but `an [access()] function` If I were translating into English, I'd need to know the text.

Also, I think the message is too hard to read:

```
Define a <0>{0}</0> function that accepts a <1>{1}</1> and returns a <2>{2}</2>. You may use the <3><4>{3}</4></3>,<5>{4}</5> (<6>{5}</6>) and <7>{6}</7> (<8>{7}</8>) modules.
```

I prefer:

```
Define a <0>process()</0> function that accepts a <1>pd.DataFrame</1> and returns a <2>pd.DataFrame</2>. You may use the <3><4>math</4></3>,<5>np</5> (<6>numpy</6>) and <7>pd</7> (<8>pandas</8>) modules.
```

File: assets/js/params/Custom/Aggregations/index.js [hunk 12]
```diff
@@ -89,7 +90,8 @@ export default class Aggregations extends React.PureComponent {
             name={`${name}[add]`}
             onClick={this.handleClickAdd}
           >
-            <i className='icon-add' /> Add
+            <i className='icon-add' />
+            <Trans id='js.params.Custom.Aggregations.addButton'>Add</Trans>
```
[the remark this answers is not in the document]
adamhooper: You marked this resolved. Why?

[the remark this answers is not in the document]
giorgio93p: Because I added the missing `{' '}`

assets/js/util/ModuleCategory.js
adamhooper: Thanks for exploring!

I see no value in an unused function -- especially since it'll always be in Git, forever :). But I'll merge anyway.

assets/js/params/File/UploadedFileSelect.js
adamhooper: Yes, this is what I'm aiming for.

assets/js/params/File/UploadApiModal.js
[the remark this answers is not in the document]
adamhooper: How many Greek programmers don't understand English? I'm assuming non-English programmers learn enough English to get by.... (Looking around, I see https://developers.facebook.com/docs/live-video-api -- Facebook API docs are translated into 17 languages, while Facebook itself is translated into 100 languages.)

Do you feel a non-technical translator will accurately transform this English into Greek (or another language)? Personally, I can't see myself asking translators to translate technical documentation. ... Then again, my goal here is to avoid wasting time; so I'll merge and then we can revisit later.

The modal is behind a "feature flag" because we haven't finished designing it. Open a Workflow with an "Upload" module; then add `#feature:uploadapi` to the URL and refresh the page. Click the new "API" button (beside the "Replace" button).

assets/js/OutputIframe.js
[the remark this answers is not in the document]
adamhooper: k, as far as I can see you've been consistent about `hoverText`. We'll see how well `hoverText` works.

assets/js/util/ModuleCategory.js
adamhooper: For module categories: let's ask them to capitalize the first letter and not the rest. I'll pester Pierre about changing the design to match. Otherwise, we need to ask translators for _both_ ... which is annoying.

Okay, let's capitalize the original text of `rows` and `columns` to `ROWS` and `COLUMNS` and nix the CSS rule.

assets/js/ImportModuleFromGitHub.js
```diff
@@ -5,6 +5,7 @@ import { updateModuleAction } from './workflow-reducer'
 import { connect } from 'react-redux'
 import { Trans, t } from '@lingui/macro'
 import { withI18n } from '@lingui/react'
+import { CategoryName } from './util/ModuleCategory'
```
adamhooper: I think the file should be `./CategoryName`

assets/js/params/Custom/Code/AceEditor.js
```diff
@@ -74,15 +75,15 @@ export default class WorkbenchAceEditor extends React.PureComponent {
...
-          (<a target='_blank' rel='noopener noreferrer' href='https://pandas.pydata.org/pandas-docs/stable/api.html#dataframe'>Pandas</a>)
-          {' '}and <kbd>np</kbd>{' '}
-          (<a target='_blank' rel='noopener noreferrer' href='https://docs.scipy.org/doc/numpy/reference/routines.html'>Numpy</a>)
-          {' '}modules.
+          <Trans id='js.params.Custom.Code.AceEditor.help' description='The tags <3>, <6>, and <8> are URLs. The rest are code formatting. Please keep code and names of libraries untranslated.'>
+              Define a <kbd>process(table)</kbd> function that accepts
+              a <kbd>pd.DataFrame</kbd> and returns
+              a <kbd>pd.DataFrame</kbd>. You may use
+              the <a target='_blank' rel='noopener noreferrer' href='https://docs.python.org/3/library/math.html'><kbd>math</kbd></a>, <kbd>pd</kbd>{' '}
+              (<a target='_blank' rel='noopener noreferrer' href='https://pandas.pydata.org/pandas-docs/stable/api.html#dataframe'>Pandas</a>) {' '}
```
adamhooper: When you put `{' '}` after a _space_, I think that produces the wrong HTML: it becomes `)  and` instead of `) and` (there's an extra space). Please don't do that.

That said, I'll merge because I hope CSS `white-space: normal;` will hide the error.

assets/js/params/File/UploadedFileSelect.js
```diff
@@ -105,11 +107,15 @@ const UploadedFileSelect = React.memo(function UploadedFileSelect ({ value, file
 
   const valueIndex = files.findIndex(({ uuid }) => uuid === value)
   const canOpen = files.findIndex(({ uuid }) => uuid !== value) !== -1 // do not open when nothing to select
+  const nFiles = files.length
+  const fileNumber = valueIndex === -1 ? <Trans id='js.params.File.UploadedFileSelect.choosePreviousFile.none'>[NONE]</Trans> : (valueIndex + 1)
```
adamhooper: Sigh, I just realized that `valueIndex + 1` won't be number-formatted when we use this pattern.

I say, ignore the bug. It may not ever affect anybody, since the numbers are always small.

assets/js/params/Custom/Code/AceEditor.js
giorgio93p: This is parsed by lingui and stored in the translation file for English, from where it is fetched to show to the user. The message in the catalog has no double space, so Lingui parses this as we intend.

PR #147, opened 2019-12-02
Support internationalized messages with quick fixes in ProcessResult.coerce

giorgio93p:


cjwkernel/types.py
```diff
@@ -547,12 +555,18 @@ def TODO_i18n(cls, text: str) -> I18nMessage:
...
     @classmethod
-    def from_dict(cls, value: Dict[str, Any]) -> I18nMessage:
+    def from_dict(cls, value: I18nMessageDict) -> I18nMessage:
         return cls(value["id"], value["arguments"])
 
-    def to_dict(self) -> Dict[str, Any]:
+    def to_dict(self) -> I18nMessageDict:
         return {"id": self.id, "arguments": self.args}
 
+    def localize(self, locale_id: str) -> str:
```
adamhooper: Nix this. We can't localize from within the `cjwkernel` package, and nobody should want to.

cjwkernel/types.py
```diff
@@ -792,6 +806,16 @@ def to_dict(self) -> Dict[str, Any]:
             "quickFixes": [qf.to_dict() for qf in self.quick_fixes],
         }
 
+    def to_pandas_error(self):
```
adamhooper: Nix this function and implement something in `cjwkernel.pandas.types` instead. `ProcessError.from_arrow()` and `ProcessError.to_arrow()` would make sense.

The `cjwkernel.types` module is (perhaps ill-advisedly) aware of Thrift; but it is not aware of Pandas.

staticmodules/scrapetable.py
```diff
@@ -77,7 +77,7 @@ def render(table, params, *, fetch_result):
     tablenum: int = params["tablenum"] - 1  # 1-based for user
 
     if tablenum < 0:
-        return ProcessResult(error="Table number must be at least 1")
+        return ProcessResult.coerce("Table number must be at least 1")
```
adamhooper: Don't act upon this comment in this pull request (please!) ... but I'd like to illustrate more about our architecture, to be as clear as possible.

This should be `return "Table number must be at least 1"`

If we're calling `ProcessResult.coerce()`, we're doing it wrong. Modules shouldn't be aware of this type.

I went through a phase a year ago in which I thought all modules would use these "pandas types" natively. But we never wired it up for external modules, and I've since learned that Pandas is incapable of doing what we want. That's why we use Arrow now.

We always had a problem with unit tests, and that's why I wavered so much. The problem with making modules return native types was, we didn't have a good way to test the results. A `render()` function can `return None`, `return pd.DataFrame()` or `return pd.DataFrame(index=pd.RangeIndex(0))` and the outcome will be identical; but in our _unit tests_, we had to get the right one. We ended up writing unit tests that catered to the module _code_, instead of the module _function_. Our expected values often wouldn't validate, and we wouldn't catch the error until deploy. That's why I took us down the `ProcessResult()` route (before finally aborting).

Lately I've solved the unit-testing problem (for now) with `cjwkernel.tests.pandas.util.assert_process_result_equal(actual, expected)`. It coerces both `actual` and `expected`.

In a future commit, we'll wipe all reference to `ProcessResult` from staticmodules. This may be all we need to do to convert them to external modules.

cjwkernel/pandas/types.py
```diff
@@ -455,7 +488,7 @@ def coerce(cls, value: Any) -> QuickFix:
                 json.dumps(args)
             except TypeError as err:
                 raise ValueError(str(err))
-            return QuickFix(text, action, args)
+            return QuickFix(_coerce_to_i18n_message_dict(text), action, args)
```
adamhooper: `return QuickFix.coerce(...)`

cjwkernel/pandas/types.py
```diff
@@ -644,15 +722,13 @@ class ProcessResult:
     modules are unreachable. Usually that means `error` should be set.
     """
 
-    error: str = ""
-    """Error (if `dataframe` is zero) or warning text."""
+    errors: ProcessResultError = field(default_factory=list)
```
adamhooper: This should not be a `moduletype`: it should be a `errors: List[RenderError] = field(default_factory=list)`. `RenderError` should be a dataclass very similar to `atypes.RenderError`.

cjwkernel/pandas/types.py
```diff
@@ -770,75 +853,119 @@ def coerce(
...
+                    if dataframe is None:
+                        dataframe = pd.DataFrame()
+                    try:
+                        errors = (
+                            []
+                            if error is None
+                            else _coerce_to_process_result_error(error)
+                        )
+                    except TypeError:
+                        errors = None
```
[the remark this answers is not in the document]
giorgio93p: I wouldn't prefer hiding the error either, but looks like `ProcessResult.coerce` never wants to raise when its argument is a tuple; it's also tested this way.

giorgio93p:
Thank you for all your comments, they were really helpful for understanding the assumptions behind the code. I think `pandas.types` looks much better now, indeed.

cjwkernel/pandas/types.py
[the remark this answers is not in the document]
adamhooper: Ah -- whoops, you're right, I forgot the contract. Indeed, we aren't meant to raise a `ValueError`: we're meant to return a `ProcessResult` with an error message.

My comment still holds: I think it would be nice if the error-`ProcessResult` (returned two lines down) included the `TypeError` text.

(And I'm still not 100% sure that's what we want. I'm comfortable merging this as-is and revisiting later if a user asks.)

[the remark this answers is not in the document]
giorgio93p: It would be a little tricky: when the `TypeError` is constructed, we don't not know where it may end up. If we construct it with a string as error, we will not be able to convert the error to `I18nMessage` in this part of `coerce` (at least without using regexps, which however seem a monstrous idea). If we construct it with an `I18nMessage` as error, we will not have a pretty error message in other parts of `coerce` where the `TypeError` is not caught (or, in fact, caught and reraised as `ValueError`).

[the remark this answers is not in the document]
adamhooper: These messages aren't i18n. While the errors can appear to end users, they are error _codes_ destined for module authors. End users won't understand them, no matter what language they're in.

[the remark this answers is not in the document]
giorgio93p: In this case, it's easy to include it. I pushed a commit and I am thus ready for a review (or merge).

cjwkernel/pandas/types.py
```diff
@@ -649,49 +702,87 @@ def arrow_table_to_dataframe(
...
+                raise ValueError(
+                    "The list %s cannot be coerced to module error" % value
+                ) from e
+
+    @classmethod
+    def coerce(cls, value: mtypes.ModuleErrorWithQuickFixes) -> ProcessResultError:
+        """Convert an error as returned by module to a member of this dataclass.
+        
+        Raises:
+        - TypeError, if the value is of incorrect type
```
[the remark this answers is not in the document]
giorgio93p: Python documentation recommends raising [`TypeError`](https://docs.python.org/2/library/exceptions.html#exceptions.TypeError) when the value is of incorrect type and [`ValueError`](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError) when the value is of correct type but invalid in some other way.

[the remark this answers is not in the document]
giorgio93p: However, I am nixing `TypeError`s from coercing, since already `ProcessResult.coerce` violates this recommendation (and I suppose it will not use `TypeError`s anytime soon).

cjwkernel/pandas/types.py
```diff
@@ -709,34 +708,24 @@ class ProcessResultError:
...
-        """Convert a list of errors as returned by module to a list of members of this dataclass.
+        """Convert a single error or a list of errors as returned by module to a list of members of this dataclass.
         
-        Raises:
-        - TypeError, if the value is not a list
-        - ValueError, if some element of the list cannot be coerced to a member of this dataclass
+        Raises ValueError, if some element of the list cannot be coerced to a member of this dataclass
         """
-        if value is None:
+        if not error_or_errors:
```
adamhooper: `if error_or_errors is None:`

Otherwise we'll do the wrong thing given `0`, `False`, or user-defined types.

cjwkernel/pandas/types.py
```diff
@@ -503,16 +502,16 @@ def coerce(cls, value: Any) -> QuickFix:
...
 
-            qf = dict(value)  # shallow copy
+            kwargs = dict(value)  # shallow copy
             try:
-                qf["text"] = I18nMessage.coerce(qf["text"])
+                kwargs["text"] = I18nMessage.coerce(kwargs["text"])
             except KeyError as err:
                 raise ValueError("Missing text from quick fix")
-            except TypeError as err:
+            except ValueError as err:
```
adamhooper: Nix?

cjwkernel/pandas/types.py
```diff
@@ -747,19 +736,19 @@ def coerce(cls, value: mtypes.ModuleErrorWithQuickFixes) -> ProcessResultError:
                 message = I18nMessage.coerce(value["message"])
             except KeyError:
                 raise ValueError("Missing 'message' in %s" % value)
-            except (TypeError, ValueError) as err:
+            except ValueError as err:
```
adamhooper: Nix?

cjwkernel/pandas/types.py
```diff
@@ -747,19 +736,19 @@ def coerce(cls, value: mtypes.ModuleErrorWithQuickFixes) -> ProcessResultError:
...
                 raise ValueError("Invalid error message") from err
 
             try:
                 quick_fixes = [QuickFix.coerce(qf) for qf in value["quickFixes"]]
             except KeyError:
                 raise ValueError("Missing 'quickFixes' in %s" % value)
-            except (TypeError, ValueError) as err:
+            except ValueError as err:
```
adamhooper: Nix?

cjwkernel/pandas/types.py
```diff
@@ -945,192 +926,125 @@ def coerce(
...
+    ) -> ProcessResult:
+        if "message" in value and "quickFixes" in value:
+            return cls(errors=[ProcessResultError.coerce(value)])
+        else:
+            value = dict(value)  # shallow copy
+            try:
+                errors = ProcessResultError.coerce_list(value.pop("errors"))
+            except KeyError:
+                errors = []
+            except ValueError as e:
```
adamhooper: Nix

cjwkernel/pandas/types.py
```diff
@@ -945,192 +926,125 @@ def coerce(
...
+                raise ValueError("Invalid 'errors' property") from e
+
+            # Coerce old-style error and quick_fixes, if it's there
+            try:
+                legacy_error_message = I18nMessage.coerce(value.pop("error"))
+            except KeyError:
+                legacy_error_message = None
+            except ValueError as e:
```
adamhooper: Nix

cjwkernel/pandas/types.py
```diff
@@ -945,192 +926,125 @@ def coerce(
...
+                errors.append(
+                    ProcessResultError(legacy_error_message, legacy_error_quick_fixes)
                 )
-            ]
-        )
+
+            dataframe = value.pop("dataframe", pd.DataFrame())
+            validate_dataframe(dataframe)
+
+            try:
```
adamhooper: With the indentation fixed, I can see another bug now: this doesn't always call `_infer_columns()`.

The fix: nix try/catch:

```
column_formats = value.pop("column_formats", {})
value["columns"] = _infer_columns(dataframe, column_formats, try_fallback_columns)
```

cjwkernel/pandas/types.py
```diff
@@ -945,192 +926,125 @@ def coerce(
...
+            except KeyError:
+                pass
+
+            try:
+                return cls(dataframe=dataframe, errors=errors, **value)
+            except TypeError as err:
+                raise ValueError(
+                    (
+                        "ProcessResult input must only contain {dataframe, "
+                        "error, errors, json, quick_fixes, column_formats} keys"
```
adamhooper: Please nix `error` and `quick_fixes` from this message. They're deprecated.

cjwkernel/pandas/types.py
giorgio93p: So `column_formats` becomes required?

cjwkernel/pandas/types.py
giorgio93p: In fact, some modules (at least `pythoncode`) may return `""` as error, so I used `if error_or_errors is None or error_or_errors is ""`

PR #150, opened 2019-12-09
Locale switcher

giorgio93p:
This pull request inserts a locale switcher in django templates and updates the locale switcher in react pages.

### DIFF
--- a/assets/js/WfHamburgerMenu.js
+++ b/assets/js/WfHamburgerMenu.js
@@ -4,7 +4,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { UncontrolledDropdown, DropdownToggle, DropdownMenu, DropdownItem } from './components/Dropdown'
 import ImportModuleFromGitHub from './ImportModuleFromGitHub'
-import LocaleSwitcher, { enableLocaleSwitcher } from './i18n/LocaleSwitcher'
+import LocaleSwitcher, { isLocaleSwitcherEnabled } from './i18n/LocaleSwitcher'
 import { Trans, t } from '@lingui/macro'
 import { withI18n } from '@lingui/react'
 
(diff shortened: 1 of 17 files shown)


cjworkbench/i18n/templates.py
```diff
@@ -8,9 +8,17 @@ def context_processor(request) -> dict:
...
 def get_i18n_context(locale=None, user=None) -> dict:
+    current_locale_id = locale or getattr(user, "locale_id", default_locale)
     return {
         "i18n": {
-            "locale_id": locale or getattr(user, "locale_id", default_locale),
+            "locale_id": current_locale_id,
             "show_switcher": getattr(user, "is_staff", False) if user else False,
+            "locales_data": [
+                {
+                    "locale_id": locale_id,
```
adamhooper: This object's keys should be `"id"` and `"name"`.

I know that in `serializers.py` we currently put understores in some JS Object keys. That's always been a mistake. (It's hard to correct it after the fact.) Let's not introduce more :).

assets/js/WfHamburgerMenu.js
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { UncontrolledDropdown, DropdownToggle, DropdownMenu, DropdownItem } from './components/Dropdown'
 import ImportModuleFromGitHub from './ImportModuleFromGitHub'
-import LocaleSwitcher from './i18n/LocaleSwitcher'
+import LocaleSwitcher, { enableLocaleSwitcher } from './i18n/LocaleSwitcher'
```
adamhooper: Could we please call this `isLocaleSwitcherEnabled`?

(`enableLocaleSwitcher` starts with a verb, so the name implies `function enableLocaleSwitcher() { isLocaleSwitcherEnabled = true }`.)

assets/js/i18n/locales.js
```diff
@@ -1,16 +1,14 @@
...
-  el: t('js.i18n.locales.el')`Greek`
-}
+export const supportedLocalesData = window && window.i18nConfig && window.i18nConfig.localesData ? window.i18nConfig.localesData : [
+  { locale_id: 'en', locale_name: 'English' }
+]
 
-export const supportedLocaleIds = Object.keys(supportedLocales)
+export const supportedLocaleIds = supportedLocalesData.map(localeData => localeData.locale_id)
 
 const defaultLocale = 'en'
```
adamhooper: I know this is a bit beyond the scope of this pull request, but I have another favor to ask:

Could these variables please be called `defaultLocaleId` and `currentLocaleId`?

I was going to make the change myself, but I figured you're already in the code ... so if I edit, it could make merges more difficult.

PR #152, opened 2019-12-10
Fix secrets info in lessons

giorgio93p:
This pull request intends to fix `analyze-live-twitter` lesson (and also deletes an unused image from a lesson).

I have not managed to test that it actually works, since creating a developer account in twitter looks too bureaucratic.

However, integration test passes on my machine...


adamhooper:
This is awesome -- thank you!

Some context:

* The build failure was my bad -- I think I committed a failing unit test.
* We don't integration-test most lessons. Integration tests are expensive to write and maintain; and we don't have a place for integration tests to store a Twitter secret.

(We integration-test the first two lessons. And that catches a ton of bugs. But the rest often fail.)

PR #153, opened 2019-12-10
First version of Greek UI translations

giorgio93p:
Also added some translator comments, based on what our translator found puzzling

### DIFF
--- a/assets/js/params/Custom/Filters/Condition.js
+++ b/assets/js/params/Custom/Filters/Condition.js
@@ -64,7 +64,7 @@ export class Condition extends React.PureComponent {
         value={value}
         onChange={this.handleChange}
       >
-        <option value=''>{i18n._(t('js.params.Custom.Filters.Condition.selectCondition')`Select condition`)}</option>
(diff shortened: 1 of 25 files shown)


assets/js/params/Custom/Filters/Condition.js
```diff
@@ -64,7 +64,7 @@ export class Condition extends React.PureComponent {
         value={value}
         onChange={this.handleChange}
       >
-        <option value=''>{i18n._(t('js.params.Custom.Filters.Condition.selectCondition')`Select condition`)}</option>
+        <option value=''>{i18n._(/* i18n: As in 'select a condition relative to which we will filter'. Condition here is something of the form 'IF something IS like that'. */t('js.params.Custom.Filters.Condition.selectCondition')`Select condition`)}</option>
```
adamhooper: A general query about this param: reading this comment, I realize the ordering of form elements -- `IF`, `[columns]`, `[behaves like]`, `[value]` -- may be ... English. Is that the case? Does the form design itself make sense in Greek?

giorgio93p: It makes complete sense in Greek

PR #156, opened 2019-12-11
Update i18n middleware and view to use locale_id in user profile

giorgio93p:
This pull request incorporates the addition of `locale_id` to `UserProfile` into i18n middleware and locale changing view.

It is not ready yet: integration tests fail because the users added in integration test helpers do not have a profile. @adamhooper, I would like some advice on it.


adamhooper:
D'oh -- that's an error in the integration-test suite. I'll update the test suite very soon; then you can rebase against master.

We should actually use integration tests to test language selection, in `integrationtests/test_I18n.py`. But I'm okay with deploying first and integration-testing later.

giorgio93p:
Thank you

I have already added a todo in my personal list for an integration test for locale switching. I haven't done it because we didn't have a final (as final it can be!) design for the switcher.

adamhooper:
@giorgio93p I just pushed commit 98bc4c8c6598f390a544b3beb58bcf106331d6e1 which adds a UserProfile for every user in integration tests. (We should already assume every user has a UserProfile on production.)

giorgio93p:
Although in principle not related to this pull request, I also pushed a commit that shows each locale in its own language.

giorgio93p:
Integration test passes on my machine; I don't have access rights to see where it falied in Google Container Builder. It's not the first time this happens.

adamhooper:
`bin/dev unittest`

cjworkbench/middleware/i18n.py
```diff
@@ -64,11 +64,11 @@ def _get_locale_from_current_user(self, request):
             locale = request.session.get("locale_id")
             return locale if is_supported(locale) else None
         else:
-            locale = getattr(request.user, "locale_id", None)
+            locale = getattr(request.user.user_profile, "locale_id", None)
             return locale if is_supported(locale) else None
 
     def _use_session(self, user):
-        return not user.is_authenticated or not hasattr(user, "locale_id")
+        return not user.is_authenticated or not hasattr(user.user_profile, "locale_id")
```
adamhooper: When does `user_profile` not have a `locale_id` attribute?

cjworkbench/i18n/views.py
```diff
@@ -23,4 +23,7 @@ def set_locale(request):
     locale = request.POST.get("new_locale")
     if is_supported(locale):
         request.session["locale_id"] = locale
+        if request.user.is_authenticated:
+            request.user.user_profile.locale_id = locale
+            request.user.user_profile.save()
```
adamhooper: Never call `Model.save()` without arguments. It's race-prone.

Use `request.user.user_profile.save(update_fields=['locale_id'])`

(Django ORM slogan: "make easy tasks hard and hard tasks impossible.")

cjworkbench/tests/test_i18n_middleware.py
```diff
@@ -15,11 +15,15 @@ def mock_response(request):
     return response
 
 
+class MockUserProfile:
+    def __init__(self, locale_preference=None):
+        self.locale_id = locale_preference
+
+
 class MockUser:
     def __init__(self, locale_preference=None):
```
adamhooper: Call the param `user_profile__locale_id`

cjworkbench/tests/test_i18n_middleware.py
```diff
@@ -15,11 +15,15 @@ def mock_response(request):
     return response
 
 
+class MockUserProfile:
+    def __init__(self, locale_preference=None):
```
adamhooper: Call this param `locale_id`.

adamhooper:
@giorgio93p Since it's Friday night, I won't merge this. But you're welcome to merge it Monday morning before I get in.

PR #157, opened 2019-12-12
Message extraction for module spec file

giorgio93p:
This pull requests adds a new command that extracts i18n messages from module spec files.

Given that the existing merging script for UI messages was not tested and that much of its functionality was also needed here, I refactored it and added unit tests. Edit: this is now [pull request 158](https://github.com/CJWorkbench/cjworkbench/pull/158).

If the module message files do not exist, the script creates them (and any parent folders it needs to), however this sets `root` as file owner. I don't know how (if) I can solve this


cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
+from cjworkbench.i18n import supported_locales, default_locale
```
adamhooper: Could we please call the module `i18n`?

Underscores in module names are un-Pythonic. I know we have a few. (The most notable is `module_loader`, which really should be `loader`). But we don't need more.

cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
...
+    old_catalog: Catalog, source_catalog: Catalog, fuzzy: FrozenSet[MessageUID]
+) -> Catalog:
+    catalog = Catalog(locale_id)
+    fill_catalog(catalog, source_catalog, old_catalog)
+    mark_fuzzy(catalog, fuzzy, old_catalog)
+    return catalog
+
+
+def _find_spec_messages(spec: ModuleSpec) -> Catalog:
+    prefix = f"modules.{spec.id_name}._spec"
```
adamhooper: Let's nix the `prefix` var and just begin every key with `_spec` (which has one fewer character than `prefix`).

I don't think we want to start every i18n key with `modules.{id_name}.` Is there a good reason to have that longer prefix?

cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
...
+    messages = {
+        f"{prefix}.name": spec.get("name"),
+        f"{prefix}.description": spec.get("description"),
+        f"{prefix}.row_action_menu_entry_title": spec.get(
+            "row_action_menu_entry_title"
+        ),
+        f"{prefix}.deprecated.message": spec.get("deprecated", {}).get("message"),
+    }
+    for param in spec.parameters:
```
adamhooper: I think this is too low a level of abstraction. Params are actually rather complex, and we already have tools to parse them.

Should be:

```
from cjwstate.models.param_spec import ParamSpec
for param_dict in spec.parameters:
    param_spec = ParamSpec.from_dict(param_dict)
    messages.update(_find_param_messages(param_spec, param_spec.id_name))
```

cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
...
+        messages = {
+            **_find_param_messages(param, f"{prefix}.parameters.{param['id_name']}"),
+            **messages,
+        }
+    return messages
+
+
+def _find_param_messages(param, param_prefix):
```
adamhooper: First off: please declare types here. I was expecting a `ParamSpec`, but this looks like a `Dict`?

Second: the correct way to visit trees is the visitor design pattern. We use Python `@singledispatch` for this. Look to `renderprep.py` for an example. (renderprep uses ParamDType, but the same pattern works for ParamSpec.) Every type is listed in `cjwstate/models/param_spec.py`. Keep in mind that optional-string fields usually have empty strings: we don't want to translate empty strings. (We assume the module author designed empty string for a reason.)

So something like:

```
def _add_if_set(l: List[Tuple[str, str]], spec: ParamSpec, prefix: str, key: str) -> None:
    value = spec.getattr(key)
    if value:
        l.append((f"{prefix}.{key}", value)


@singledispatch
def extract_param_messages(spec: ParamSpec, prefix: str) -> List[Tuple[str, str]]:
    return []  # fallback for types we don't handle


@extract_param_messages.register(ParamSpec.String)
def _(spec: ParamSpec.String, prefix: str) -> List[Tuple[str, str]]:
    ret = []
    _add_if_set(ret, spec, prefix, "name")
    _add_if_set(ret, spec, prefix, "placeholder")
    _add_if_set(ret, spec, prefix, "default")
    return ret

@extract_param_messages.register(ParamSpec.Secret)
def _(...)
```

... this ends up being rather verbose, so I recommend putting it in a separate Python module.

The win is: correctness. Treat `param_spec.py` as a to-do list: it lists all types and all fields for those types.

cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
...
+    messages = {}
```
adamhooper: My example is `messages = []` ... but I'm not strongly opinionated about it.

server/management/commands/extract-module-messages.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def main(directory):
+    cjwstate.modules.init_module_system()
+
+    logger.info(f"Extracting...")
+
+    with tempfile.TemporaryDirectory() as tmpdir:
```
adamhooper: This tempfile stuff isn't needed at all. (We use it in `develop-module` because `import_module_from_directory()` copies all files, slowing down development significantly.)

cjwstate/modules/message_catalogs.py
```diff
@@ -0,0 +1,116 @@
...
+            old_catalog = read_po_catalog(po_path)
+        except FileNotFoundError:
+            old_catalog = Catalog(locale_id)
+        write_po_catalog(
+            po_path, _merge_nonsource_catalog(old_catalog, source_catalog, fuzzy)
+        )
+
+
+def _po_path(basepath: os.path, locale_id: str, module_id_name: str) -> os.path:
+    return os.path.join(basepath, "locale", locale_id, f"{module_id_name}.po")
```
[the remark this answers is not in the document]
giorgio93p: I tried to make it easy to support staticmodules, shall we decide to keep them. Nixing the name of the module both from the filename and from the message IDs (as suggested in another one of your comments) would be a problem for them.

cjworkbench/i18n/catalogs/util.py
```diff
@@ -104,10 +135,14 @@ def read_po_catalog(filename: str) -> Catalog:
         return read_po(catalog_file)
 
 
-def write_po_catalog(filename: str, catalog: Catalog, **kwargs):
+def write_po_catalog(filename: Union[str, pathlib.Path], catalog: Catalog, **kwargs):
```
adamhooper: In our own internal APIs, permissive function signatures aren't worth the risk. Please nix the `Union` and be opinionated about type. (`pathlib.Path` is the one I'd pick, and I'd call the argument `path`.)

cjwstate/modules/i18n/__init__.py
```diff
@@ -0,0 +1 @@
+from cjwstate.modules.i18n.catalogs import update_module_catalogs
```
adamhooper: To me, it seems like `update_module_catalogs()` does for modules what `bin/dev extract-messages` does for our core code. Should this be called `extract_module_messages()`? (Is the concept different enough to warrant a new verb?)

cjwstate/modules/i18n/catalogs.py
```diff
@@ -0,0 +1,75 @@
...
+    source_catalog = _build_source_catalog(ModuleSpec.load_from_path(module_files.spec))
+
+    po_path = _po_path(directory, default_locale)
+
+    try:
+        old_source_catalog = read_po_catalog(po_path)
+    except FileNotFoundError:
+        old_source_catalog = Catalog(default_locale)
+
+    if not catalogs_are_same(source_catalog, old_source_catalog):
```
adamhooper: I'm geeking out at how legible this code is :).

What happens in this scenario:

1. Module author publishes a module in English and French
2. Workbench enables Portuguese
3. Module author runs `update_module_catalogs()`

Expected results: this function writes a new `<module>/locale/pt/messages.po`
Actual results: it doesn't, because `<module>/messages.pot` is unchanged

adamhooper:
Whoops. @giorgio93p could you please `git merge master`? Then I'll merge.

PR #158, opened 2019-12-13
Refactor catalog merge scripts and unit test them

giorgio93p:
This pull request singles out some changes of [https://github.com/CJWorkbench/cjworkbench/pull/157](https://github.com/CJWorkbench/cjworkbench/pull/157)

### DIFF
--- a/cjworkbench/tests/test_catalogs.py
+++ b/cjworkbench/tests/test_catalogs.py
@@ -1,0 +1,848 @@
+from django.test import SimpleTestCase
+from babel.messages.catalog import Catalog, Message
+from cjworkbench.i18n.catalogs.util import (
+    message_unique_identifier,
+    find_corresponding_message,
+    find_corresponding_string,
+    find_fuzzy_messages,
+    fill_catalog,
+    mark_fuzzy,
+    remove_strings,
+)
+from cjworkbench.i18n.catalogs.merge import _merge_source_catalog, _merge_catalog
+
+
+class CatalogTest(SimpleTestCase):
+    def assertCatalogsDeeplyEqual(self, catalog: Catalog, other_catalog: Catalog):
+        """ Assert that the two catalogs contain the same messages, where message equality is deep.
+        
+        Ignores header message.
+        """
+        self.assertCatalogDeeplyContainsCatalog(catalog, other_catalog)
+        self.assertCatalogDeeplyContainsCatalog(other_catalog, catalog)
+
+    def assertCatalogDeeplyContainsCatalog(
+        self, catalog: Catalog, other_catalog: Catalog
+    ):
+        """ Assert that `other_catalog` contains all the messages of `catalog`, where message equality is deep.
+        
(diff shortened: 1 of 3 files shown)


cjworkbench/tests/test_catalogs.py
```diff
@@ -0,0 +1,848 @@
...
+    def assertMessageDeeplyEqual(self, message: Message, other_message: Message):
+        self.assertEqual(message, other_message)  # this compares id and context
+        self.assertEqual(message.string, other_message.string)
+        self.assertEqual(message.flags, other_message.flags)
+        self.assertEqual(message.auto_comments, other_message.auto_comments)
+        self.assertEqual(message.user_comments, other_message.user_comments)
+        self.assertEqual(message.locations, other_message.locations)
+
+
+class UtilTest(CatalogTest):
```
adamhooper: Why is this a `SimpleTestCase` instead of a plain `unittest.TestCase`?

giorgio93p: No idea. I just copied from some other test file.

cjworkbench/tests/test_catalogs.py
```diff
@@ -0,0 +1,848 @@
...
+    find_corresponding_string,
+    find_fuzzy_messages,
+    fill_catalog,
+    mark_fuzzy,
+    remove_strings,
+)
+from cjworkbench.i18n.catalogs.merge import _merge_source_catalog, _merge_catalog
+
+
+class CatalogTest(SimpleTestCase):
```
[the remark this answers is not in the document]
giorgio93p: It doesn't need to, but I like `self.assert...` more than `assert_...`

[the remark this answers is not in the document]
adamhooper: I'd prefer `assert_...()`, please. We already use that pattern with Numpy, Pandas, Arrow, ProcessResult.

Someday, with any luck, we'll abandon the `unittest` module altogether. All the cool kids are using pytest, because it's better in every way.

cjworkbench/tests/test_catalogs.py
adamhooper: Please extend `unittest.TestCase`. It has fewer features. It doesn't depend on Django. (In case it isn't comically clear: I _hate_ depending on Django ;).)

adamhooper:
@giorgio93p Oh, also, I just noticed: please rename `tests/test_catalogs.py` to `tests/i18n/catalogs/test_merge.py` and call the class `MergeTest`.

giorgio93p:
`tests/test_catalogs.py` tests two different file, so I guess I should split it to  `tests/i18n/catalogs/test_merge.py`, `tests/i18n/catalogs/test_util.py`, and a third file (for which the name `tests/i18n/catalogs/util.py` is probably fine, but I am not too sure) which will contain `assert_*`

adamhooper:
Yeah, that sounds perfect :)

PR #161, opened 2019-12-14
Change how javascript expects renderring errors

giorgio93p:
Javascript expects errors to already have been translated by the backend.
It also mirrors the structure of RenderErrors: a list whose elements contain an error (string) and a list of quick fixes.

Notice that the dataclasses in `cjwkernel.types` temporarily have `localize` methods. These will be removed when we use the new Deltas, which will also handle localization.


assets/js/WorkflowEditor/wfmodule/StatusLine.js
```diff
@@ -46,7 +46,7 @@ const StatusLine = React.memo(function StatusLine ({ status, errors, applyQuickF
 StatusLine.propTypes = {
   status: PropTypes.oneOf(['ok', 'busy', 'error', 'unreachable']).isRequired,
   isReadOnly: PropTypes.bool.isRequired, // if true, cannot apply quick fixes
-  errors: PropTypes.arrayOf(PropTypes.shape({ message: PropTypes.string.isRequired, quickFixes: PropTypes.arrayOf(PropTypes.shape(QuickFixPropTypes)) })), // may be empty
+  errors: PropTypes.arrayOf(PropTypes.shape({ message: PropTypes.string.isRequired, quickFixes: PropTypes.arrayOf(PropTypes.shape(QuickFixPropTypes)).isRequired }).isRequired).isRequired, // may be empty
```
adamhooper: Please split this onto multiple lines. It's _still_ missing an `.isRequired`. (`PropTypes.shape(QuickFixPropTypes).isRequired`)

```
errors: PropTypes.arrayOf(
    PropTypes.shape({
        message: PropTypes.string.isRequired,
        quickFixes: PropTypes.arrayOf(PropTypes.shape(QuickFixPropTypes).isRequired).isRequired
    }).isRequired
).isRequired
```

assets/js/params/Custom/Code/AceEditor.js
```diff
@@ -17,7 +18,7 @@ export default class WorkbenchAceEditor extends React.PureComponent {
   static propTypes = {
     // When isZenMode changes, we'll call componentDidUpdate()
     isZenMode: PropTypes.bool.isRequired,
-    wfModuleOutputError: PropTypes.string, // hopefully null or empty
+    wfModuleOutputErrors: PropTypes.arrayOf(PropTypes.shape({ message: PropTypes.string.isRequired, quickFixes: PropTypes.arrayOf(PropTypes.shape(QuickFixPropTypes)).isRequired }).isRequired).isRequired, // may (hopefully) be empty
```
adamhooper: Gah, now we're reusing this big block that has an error :).

I'll merge this and then refactor the prop types. (QuickFixPropTypes follows a bad convention, for a bad reason.) I should have done this ages ago.

renderer/execute/types.py
```diff
@@ -122,38 +155,14 @@ def as_error_message(self):
...
         super().__init__("user must change something before we render")
         self.errors = errors
 
     def __eq__(self, other):
         return isinstance(other, self.type) and other.errors == self.errors
 
-    def as_errors_with_fixes(self) -> List[Tuple[I18nMessage, List[QuickFix]]]:
-        return [(err.as_error_message(), err.as_quick_fixes()) for err in self.errors]
+    def as_render_errors(self) -> List[RenderError]:
+        return reduce(
```
adamhooper: I love functional programming as much as anybody, but in the future I suggest avoiding `reduce()`. It just doesn't work very well in Python.

Yours was:

```
return reduce(
    lambda errors, error: errors + error.as_render_errors(), self.errors, []
)
```

I suggest what I consider the Pythonic approach:

```
ret = []
for error in self.errors:
    ret.extend(error.as_render_errors())
return ret
```

Mine is 1 line longer, 10-20 chars shorter, faster and more legible.

(Don't get me wrong: I _love_ `reduce()` normally. But I have yet to find a good fit for it in Python.)

Another experiment:

```
return sum([err.as_render_errors() for err in self.errors], [])
```

This is 2 lines shorter than yours, 24 chars shorter, and equivalent. Though I suspect the average Python programmer wouldn't understand it -- just like the average Python programmer wouldn't understand `reduce()`.

I'm not asking you to change anything -- your code is correct. Just an observation for next time....

PR #162, opened 2019-12-17
Extract i18n messages from module code

giorgio93p:
This pull request adds the ability to extract messages from module python code. The developer still needs to run the extraction command (its integration in `bin/dev develop-module` will be addressed in a future pull request).

An interesting question how we expect module authors to call the `cjwmodule.i18n.trans` translation helper: just `trans`, `i18n.trans`, or any of these? For ease of implementation (the tokenizer splits on `.`), I adopted the last option: both `trans` and `i18n.trans` will work the same. This is reflected in the unit tests: most use `i18n.trans`, but `trans` is also tested.


cjwstate/modules/i18n/catalogs/extract/code.py
```diff
@@ -0,0 +1,212 @@
...
+        else:
+            messages[message_id] = {
+                "string": default_message,
+                "comments": comments,
+                "locations": [(relative_path_name, lineno)],
+            }
+    return messages
+
+
+def _extract_module_code(fileobj, keywords, comment_tags, options):
```
adamhooper: How is this different from `cjworkbench.i18n.catalogs.extract.extract_python()`?

Should the two functions be merged? It seems to me they should, but you may be thinking of something I'm missing.

giorgio93p: Indeed, most of the code was the same (parsing logic copied by `pybabel`). I changed the structure a bit, in order to avoid code duplication (and at the same time highlight the difference between the two): now the common part expects a callback to run when a translation function has been found.

cjworkbench/i18n/catalogs/extract.py
```diff
@@ -116,6 +178,11 @@ def extract_python(fileobj, keywords, comment_tags, options):
     :param comment_tags: a list of translator tags to search for and include
                          in the results
     :param options: a dictionary of additional options (optional)
+    :param process: a callable that accepts funcname, messages, translator_comments, message_lineno
```
adamhooper: Why a callback? Wouldn't it be more Pythonic for the caller to use a simple for-loop:

```
for message_lineno, funcname, messages, translator_comments in _parse_and_extract_python(...):
    if ...:
        yield funcname, messages, translator_comments, message_lineno
```

Also, with this approach, we could guarantee `extract_python()` won't return `funcname=None`.

cjworkbench/i18n/catalogs/extract.py
```diff
@@ -101,13 +101,75 @@ def extract_django(fileobj, keywords, comment_tags, options):
...
+            else:
+                funcname = None
+        return funcname, messages, translator_comments, message_lineno
+
+    return _parse_and_extract_python(fileobj, keywords, comment_tags, options, process)
+
+
+def extract_module_code(fileobj, keywords, comment_tags, options):
+    def process(funcname, messages, translator_comments, message_lineno):
+        if funcname == "trans":
```
adamhooper: Isn't this what the `keywords` argument is for?

I would expect our signature to be: `def extract_module_code(fileobj): ...`

It would call `_parse_and_extract_python(fileobj, {"trans"}, {"i18n"})`

... then in this callback, it would be impossible for `funcname` to be anything other than `"trans"`, right?

cjwstate/modules/i18n/catalogs/extract/code.py
```diff
@@ -23,7 +21,7 @@ def _find_messages_in_module_code(
     code_file, relative_path_name: str
 ) -> Dict[str, Dict[str, Any]]:
     messages_data = extract(
-        _extract_module_code,
+        extract_module_code,
         code_file,
         keywords={"trans": (1, 2)},
```
adamhooper: What does the `(1, 2)` mean?

giorgio93p: Means that we are interested in the arguments `1` and `2` of `trans`

cjworkbench/i18n/catalogs/extract.py
giorgio93p: Regarding `funcname`, you are correct

cjworkbench/i18n/catalogs/extract.py
giorgio93p: The `funcname` returned by `extract_python` may need to be `None`. `pybabel` expects only `gettext`-kind function names.

giorgio93p:
Well, the signatures of our public functions can't change, since they are called by `pybabel` internal code. I added type hints and used `yield` instead of callbacks.

cjworkbench/i18n/catalogs/extract.py
adamhooper: Ohhhhhh, gotcha.

cjwstate/modules/i18n/catalogs/extract/code.py
adamhooper: I've grepped for `keywords.get`, `keywords.items` and `keywords[` and found nothing. I can't find any place where we read values from `keywords`. Also, pybabel's Extract `keywords` argument is not a dict.

Can't it just be an `Iterable`? (I imagine passing `Set`/`FrozenSet` would be most efficient; and pybabel would probably pass `List`....)

adamhooper:
D'oh, good point re the signatures! Thanks.

adamhooper:
This is easily good enough to merge. But I'd love if you could double-check that `keywords` type.

cjwstate/modules/i18n/catalogs/extract/code.py
giorgio93p: Indeed, I copied it from the call to the higher-level `extract`, which expects a `dict`, but at this point a `List` is expected (and anything that supports the `in` operator would work).

Issue #149, opened 2019-12-09
Auth messages aren't translated to other languages


adamhooper:
Steps to reproduce:

1. Create a new account using your main email address
2. Log out
3. Log in via Google account -- on same email address

Expected results: `templates/socialaccount/messages/account_connected.txt` is translated into your language
Actual results: `context["i18n"]` is not set when `trans_html` is called.

I added a workaround to `trans_html()`: output the default language. This lowers the severity of the problem (it no longer causes a 500 error); but it's not a fix.

My hypothesis: the problem is in middleware. `cjworkbench.middleware.i18n.SetCurrentLocaleMiddleware` comes _after_ `django.contrib.auth.middleware.AuthenticationMiddleware`. That's broadly correct because we can't determine the user's locale before we've looked up the user. But in this one scenario, `i18n` isn't set because we haven't logged the user in yet when we render the message.

Brainstorming a fix: how about we split our logic into two pieces of middleware:

1. `cjworkbench.middleware.i18n.SetCurrentLocaleFromRequestAndSessionMiddleware` (comes soon after `django.contrib.sessions.middleware.SessionMiddleware`
2. `cjworkbench.middleware.i18n.SetCurrentLocaleFromLoggedInUserMiddleware` (comes soon after `django.contrib.auth.middleware.AuthenticationMiddleware`)

@giorgio93p could you please take a look?

giorgio93p:
I am trying to add Google login in my dev environment, but with no success. As per the [documentation](https://github.com/CJWorkbench/cjworkbench/wiki/Setting-up-a-development-environment#to-test-social-account-logins), I added `socialaccounts_secrets.json` and restarted docker containers, but it doesn't show up.

adamhooper:
Thanks for checking. I likely need to update the docs. (I didn't actually replicate this behavior; I'm just going by the notifications on prod.)

So 1) I'll update the docs; 3) I'll ping you :)

adamhooper:
@giorgio93p I've fixed social-account login from dev machines. The new instructions are at https://github.com/CJWorkbench/cjworkbench/wiki/Setting-up-a-development-environment#to-test-social-account-logins

giorgio93p:
I still can't see the social login buttons.
Maybe I am not doing something correctly in Google Console; it's a huge mess...
How can I see where social login fails to be set?

adamhooper:
@giorgio93p whoops -- I forgot to push! Try now

[...]

PR #163, opened 2019-12-18
Fix i18n template context in social login confirmation

giorgio93p:
Solves #149

### DIFF
--- a/cjworkbench/allauth_account_adapter.py
+++ b/cjworkbench/allauth_account_adapter.py
@@ -46,7 +46,7 @@ class AccountAdapter(DefaultAccountAdapter):
             try:
                 if message_context is None:
                     message_context = {}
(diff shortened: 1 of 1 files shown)


cjworkbench/allauth_account_adapter.py
```diff
@@ -30,3 +34,21 @@ def send_confirmation_mail(self, request, emailconfirmation, signup):
...
+    ):
+        """
+        Wrapper of `django.contrib.messages.add_message`, that reads
+        the message text from a template.
+        """
+        if "django.contrib.messages" in settings.INSTALLED_APPS:
+            try:
+                if message_context is None:
+                    message_context = {}
+                message_context.update(context_processor(request))
```
adamhooper: Do not modify the input argument. Use `{**message, **context_processor(request)}` instead.

Issue #155, opened 2019-12-10
Logout destroys language setting


adamhooper:
Steps to reproduce:

1. Log in
2. Choose Greek language
3. Log out

Expected results: things are still Greek
Actual results: back to English!

This is because when the user logs out, we destroy the session.

One solution is to use a separate cookie for language. This would make our solution to #149 seem to fit in, too. On #149 I was suggested two different pieces of middleware running at two different times. Now I'm also suggesting two different cookies -- one per middleware.

giorgio93p:
It is a bit counter-intuitive, indeed.

Question: What language should we show after the following scenario?

- Set to Greek
- Login
- Set to English
- Logout

adamhooper:
Let's go with:

After step 2: the user's language (because the user stored it in the database during a previous session)

After step 4: English (because the previous page was English -- had step 3 never happened, it would be the user's language from the database)

PR #164, opened 2019-12-18
Various changes regarding locale setting and changing

giorgio93p:
1. In order to solve issue #155, locale is now stored in a cookie instead of the session
2. When a user signs up, the current locale is saved as their locale preference
3. Added integration tests for locale switching, issue #155, and number 2 of this list
3. By the way, fixed some accents in Greek translations


cjworkbench/forms/signup.py
```diff
@@ -43,3 +43,5 @@ def signup(self, request, user):
         profile, created = UserProfile.objects.get_or_create(user=user)
         profile.get_newsletter = self.cleaned_data["get_newsletter"]
         profile.save()
+        user.user_profile.locale_id = request.locale_id
```
adamhooper: `user.user_profile` should be `profile` here.

In fact, this should all be exactly one call. And I think the call should be `update_or_create()`. So how about:

```
UserProfile.objects.update_or_create(
    user=user,
    defaults={"get_newsletter": self.cleaned_data["get_newsletter"], "locale_id": request.locale_id},
)
```

cjworkbench/i18n/views.py
```diff
@@ -20,10 +21,11 @@ def set_locale(request):
         url=next, allowed_hosts={request.get_host()}, require_https=request.is_secure()
     ):
         next = "/"
+    response = HttpResponseRedirect(next)
     locale = request.POST.get("new_locale")
     if is_supported(locale):
-        request.session["locale_id"] = locale
+        response.set_cookie(COOKIE_NAME, locale)
```
adamhooper: Add `max_age=365*86400`. Otherwise the cookie disappears when the user closes the browser window.

cjworkbench/middleware/i18n.py
```diff
@@ -29,6 +33,8 @@ def __call__(self, request):
 
         # Code to be executed for each request/response after
         # the view is called.
+        if resolve(request.path_info).url_name != "set_locale":
```
adamhooper: If you _don't_ add these two lines of code, which integration tests fail? If none of them do, let's nix this line. Otherwise, let's discuss.

I don't like hard-coding a special URL in this middleware. I feel it's jumping above its layer of abstraction.

integrationtests/helpers/accounts.py
```diff
@@ -48,19 +48,30 @@ def _clear_minio():
                 raise err
 
 
-def login(browser: Browser, email: str, password: str) -> None:
+def login(
```
adamhooper: I don't think these helpers should change. Please move your locale-enhanced functions into `test_i18n.py` instead.

It's a judgment call. I know i18n is a cross-cutting concern; but that's true of almost _everything_ we integration test. Each test is more legible when _its_ cross-cutting concern doesn't impact the helpers all the _other_ test files use.

integrationtests/test_i18n.py
```diff
@@ -2,6 +2,39 @@
 from integrationtests.helpers import accounts
 
 
+_locale_names = {"en": "English", "el": "Ελληνικά"}
+
+
+class LocaleSwitcherIntegrationTest(WorkbenchBase):
```
adamhooper: Don't create base test classes. Nobody should ever import from this file, so there's no possibility this code will be reused. Just put the helpers in the test class itself.

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
         # Wait for page to load
         b.assert_element('input[name="name"][value="Untitled Workflow"]', wait=True)
         b.assert_element(".attribution .metadata", text="by Jane Doe")
+
+
+class TestLocaleSwitch(LocaleSwitcherIntegrationTest):
```
adamhooper: No need for a new class. The other one has a buggy name -- it should be `class TestI18n(WorkbenchBase)`. And all the methods in this Python module should be within that one class.

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
...
+    def setUp(self):
+        super().setUp()
+        self.user = self.account_admin.create_user(
+            "user@example.org", first_name="Jane", last_name="Doe"
+        )
+
+    def test_language_switch_in_django_templates(self):
```
adamhooper: Please mention `..._not_logged_in` in the test name.

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
...
+        # You can change from "en" to "el" and back and the UI will be translated
+        b = self.browser
+        b.visit("/")
+        b.assert_element('input[name="password"][placeholder="Password"]', wait=True)
+        b.assert_element('button[type="submit"]', text="Sign In")
+        self.switch_locale_django("el")
+        b.assert_element('input[name="password"][placeholder="Συνθηματικό"]', wait=True)
+        b.assert_element('button[type="submit"]', text="Σύνδεση")
+        self.switch_locale_django("en")
```
adamhooper: Before switching locales, navigate to another page (`b.click("[Greek for 'Sign Up']")` springs to mind) to check that the chosen language persists to the next page.

(Don't refresh the page. Selenium is prone to races; after refreshing the page, your next `b.assert_element()` may check the _previous_ page or it may check the _next_ page. If the assertion happens to query the _previous_ page, the test isn't actually testing anything. You need to test for an element that _isn't_ on the previous page to ensure the test is useful.)

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
...
+            'form.create-workflow button[type="submit"]',
+            text="Create Workflow",
+            wait=True,
+        )
+        self.switch_locale_react("en", "el")
+        b.assert_element(
+            'form.create-workflow button[type="submit"]',
+            text="Δημιουργία ροής εργασιών",
+            wait=True,
+        )
```
adamhooper: Before switching locales, navigate to another page (`b.click("[Greek for 'Lessons']")` springs to mind) to check that the chosen language persists to the next page.

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
...
+        self.switch_locale_react("el", "en")
+        b.assert_element(
+            'form.create-workflow button[type="submit"]',
+            text="Create Workflow",
+            wait=True,
+        )
+
+
+class TestLocaleOnLogin(LocaleSwitcherIntegrationTest):
```
adamhooper: No need for a new class.

integrationtests/test_i18n.py
```diff
@@ -20,3 +53,88 @@ def test_js_english_interpolation(self):
...
+        b.assert_element('button[type="submit"]', text="Σύνδεση")
+
+        # Switch to English and then login. After login, page must be in Greek
+        self.switch_locale_django("en")
+        accounts.login(b, self.user.email, self.user.email, "en", "el")
+        b.assert_element(
+            'form.create-workflow button[type="submit"]',
+            text="Δημιουργία ροής εργασιών",
+            wait=True,
+        )
```
adamhooper: I'm not usually a fan of a test that tests three things ... but I'm going to recommend a fourth here :).

What language does the user get after logout? (IIRC we agreed it should be Greek.)

integrationtests/test_signup.py
```diff
@@ -1,7 +1,8 @@
 from integrationtests.utils import WorkbenchBase, find_url_in_email
+from integrationtests.test_i18n import LocaleSwitcherIntegrationTest
```
adamhooper: Never do this. Just don't import it.

integrationtests/test_signup.py
```diff
@@ -35,3 +36,40 @@ def test_signup(self):
         b.fill_in("password", "?P455W0rd!")
         b.click_button("Sign In")
         b.wait_for_element("a", text="MY WORKFLOWS", wait=True)
+
+    def test_signup_el(self):
```
adamhooper: Put this in `test_i18n.py`.

Otherwise, we set a precedent: all integration-test files will could test i18n. Soon enough we get into i18n-specific minutae in each file: 'which languages should we test "Example Workflows" in?' etc.

That precedent isn't foolish; but I don't think it's the best cost/benefit for us. I prefer to let `test_i18n.py` grow as we encounter i18n-related regressions.

integrationtests/test_i18n.py
giorgio93p: Logout is tested above; as mentioned in the comment, `Switch to Greek and then logout. Page must still be in Greek`.

cjworkbench/middleware/i18n.py
giorgio93p: The issue was that the `locale` decided by this middleware is to be changed by the view.

Now, the middleware does not store this `locale` in the cookie; it stores `request.locale_id`, which has been changed by the view of `set_locale`. The thing to keep in mind is that this can be also done by other views.

cjworkbench/forms/signup.py
giorgio93p: For some very weird reason, this always sets the default locale. 
That's why I used `user.user_profile` in the first place: I had tried 
```
profile = UserProfile.objects.update_or_create(user=user)
profile.locale_id = request.locale_id
profile.get_newsletter = self.cleaned_data["get_newsletter"]
profile.save()
```
and it always set the default locale too.

adamhooper: "Very weird reason" is a red flag. We should not merge very weird stuff. We should debug, come up with a simple explanation, and put that in a comment.

Here's my theory: we actually create UserProfile in two places:

* `cjworkbench/models/userprofile.py`
* `cjworkbench/forms/signup.py`

... and maybe something along the lines of: allauth is creating the user; and then our custom signal is creating the user _with the default locale_; then the user is signing in and middleware sets `request.locale_id` to user_profile.id; and _then_ we call `signup()`. (This explanation doesn't agree with the facts you've we've laid out; but then, we're in "very weird" territory so please indulge me the brainstorm... :).)

Ideally, we'd track down every call of `django.contrib.auth.models.UserManager.create_user()`. and make sure we create a UserProfile at the same time. (Then we'd nix the signal in `cjworkbench/models/userprofile.py`.) Who calls `create_user()`? I can't find it when I grep Django or Allauth code.

After we figure things out, I'd probably be fine with leaving the user_profile in signup.py. We could read the locale from the request _cookie_ (not the request itself) and write _that_ to the user profile.

cjworkbench/middleware/i18n.py
adamhooper: I think we're talking over each other here. I'll rephrase with a very specific question:

How does `request.locale_id` influence the logic in `/set_locale`?

integrationtests/test_i18n.py
adamhooper: I was talking about testing something different. What I'm asking to test is:

1. Create account that has Greek
2. Browse in English
3. Log in [should now be Greek] (does something we haven't integration-tested)
4. Log out [should now be Greek] <-- this is the test I'm requesting

This is different from what's already in the test, which is:

1. Create account
2. Log in
3. Set to Greek (we know, through integration tests, that this sets both COOKIE _and_ user-profile language)
4. Log out [should now be Greek] <-- this is what you've already tested

Since this confused me -- and our conversation wasn't seamless -- could you please create this new test separate from the existing one that tests three things?

cjworkbench/middleware/i18n.py
giorgio93p: 1. The first part of the middleware runs and looks for the current locale. Saves it in its `locale` local variable, `request.locale_id`, and django.
2. `set_locale` runs. Finds the new locale and saves it somewhere (currently, in `request.locale_id` and in user profile for registered users, formerly, in the cookie instead of `request.locale_id`).
3. The second part of the middleware runs. This must set a cookie. If it sets the cookie using its own `locale` local variable, it will undo (some of) the changes in `set_locale` view (that's why I excluded this url in the first version of the code). If it sets the cookie using `request.locale_id`, it will allow views to change the cookie indirectly.

integrationtests/test_i18n.py
giorgio93p: Let's see if I understand correctly.

Current test:
1. Create account
2. Log in
3. Set to Greek. Notice that now I have created a user with Greek locale in preferences.
4. Logout (must be in Greek)
5. Change to English
6. Login (must be in Greek)
7. Logout (must be in Greek)

I should split it in two
Test logout:
1. Create account
2. Log in
3. Set to Greek.
4. Logout (must be in Greek)

Test login:
1. Create account with Greek locale in preferences
2. Login from English page (must go to Greek page)
3. Logout (must be in Greek)

adamhooper: First, please leave the current test as-is and add a new test.

My point is, the sequence of events I asked to test is subtly different from the sequence that is already tested. This matters because elsewhere in my review of this pull request I'm suggesting you stop writing a cookie :). I worry that not-writing the cookie will not cause a failure in your test but may cause a failure in the test I'm proposing.

cjworkbench/middleware/i18n.py
adamhooper: And if the middleware _doesn't_ set a cookie? What happens?

I would expect `set_locale` and login to set a cookie. I wouldn't expect middleware to set a cookie.

giorgio93p:
This pull request has a very strange bug in integration tests: when all tests are run sequentially, `test_notifications` fails at `message.get_payload()[1].get_payload()` with error `AttributeError: 'str' object has no attribute 'get_payload'` (looks like another email is fetched). When the same test is run on its own, it passes.

cjworkbench/forms/signup.py
giorgio93p: Hmmm... Turns out it can work as intended when using `profile` instead of `user.user_profile`. Perhaps I had forgotten to add `update_fields` in `profile.save()` when I was experimenting.

integrationtests/test_notifications.py
```diff
@@ -54,7 +54,15 @@ def got_notification_email(_unused):
             if not message:
                 return False
             # Get payload at index 1: the text/html payload
-            body = message.get_payload()[1].get_payload()
+            #
+            # There is a chance that an old signup message will be fetched
+            # before the new notification is written to disk;
+            # the signup message will have a `str` at index 1, resulting in
+            # an `AttributeError` for the second `get_payload`
```
adamhooper: You figured it out! Thank you! (I'd seen this error under certain conditions and never figured it out.)

templates/account/login.html
```diff
@@ -24,7 +24,7 @@
         {% include "socialaccount/snippets/login_extra.html" %}
 
         <div class="d-flex justify-content-center mb-4">
-          <div class="content-2">or</div>
+          <div class="content-2">{% trans_html "py.templates.account.login.socialSimpleSeparator" default="or" comment="As in 'Login with Google account or login with simple account'" %}</div>
```
adamhooper: I didn't understand this comment when I saw it in the .pofiles. Can we try iterating? My brainstorm idea:

`"The 'or' in 'Log in with [Google/Facebook buttons] *or* [username/password]'"`

cjworkbench/apps.py
```diff
@@ -6,4 +6,4 @@ class ServerConfig(AppConfig):
     name = "cjworkbench"
 
     def ready(self):
-        pass
+        import cjworkbench.signals
```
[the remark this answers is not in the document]
giorgio93p: It sets up the signals to be used

assets/js/WfHamburgerMenu.js
```diff
@@ -55,7 +55,7 @@ export class WfHamburgerMenu extends React.Component {
           </DropdownToggle>
 
           <DropdownMenu>
-            <DropdownItem onClick={this.handleClickOpenLocaleSwitcher}><i className='icon icon-language' /><Trans id='js.WfHamburgerMenu.menu.language'>Language</Trans></DropdownItem>
+            <DropdownItem className='locale-switcher-show' onClick={this.handleClickOpenLocaleSwitcher}><i className='icon icon-language' /><Trans id='js.WfHamburgerMenu.menu.language'>Language</Trans></DropdownItem>
```
adamhooper: Can't the integration test click on `.dropdown-item i.icon-language`?

If we absolutely need to add a class name to make integration tests work, then we need a comment explaining why. Otherwise, the reader will assume this class is for CSS.

assets/js/i18n/LocaleSwitcher.js
```diff
@@ -57,7 +57,7 @@ export default function LocaleSwitcher ({ closeModal }) {
         </form>
       </ModalBody>
       <ModalFooter>
-        <button className='action-button button-gray' onClick={closeModal}>
+        <button className='action-button button-gray close-button' onClick={closeModal}>
```
adamhooper: The user doesn't need to click "Close" to switch locales. Clicking the language refreshes the page.

Please nix this class. As before: class names should be for CSS.

integrationtests/test_i18n.py
```diff
@@ -2,38 +2,47 @@
...
+def switch_locale_django(browser: Browser, to_locale_name: str):
+    browser.click_whatever("a#locale-switcher-dropdown", wait=True)
+    browser.click_button(to_locale_name, wait=True)
+
+    # check that the language has indeed changed
+    browser.wait_for_element("a#locale-switcher-dropdown", text=to_locale_name)
+
+
+def switch_locale_react(browser: Browser, to_locale_name: str):
+    browser.click_whatever(".navbar .dropdown button", wait=True)
```
adamhooper: Why is there a `wait=True` here? Do we ever call `switch_locale_react()` when the hamburger-menu button hasn't loaded?

With `wait=True`, the sentence is: "Wait for the hamburger-menu button to appear, then click it."

Without, the sentence is: "Click the hamburger-menu button."

I think `wait=True` should be nixed. Otherwise, we certainly need a comment explaining why we're waiting. (Our standard practice is to _always_ put a comment when using `wait=True`; but I feel the next two lines in this function don't need comments because it's perfectly clear what we're waiting for.)

integrationtests/test_i18n.py
```diff
@@ -2,38 +2,47 @@
...
+    browser.click_whatever(".locale-switcher-show", wait=True)
+    browser.click_button(to_locale_name, wait=True)
+
+    # check that the language has indeed changed
```
adamhooper: I have two issues with these next four lines:

1. They assert success. That blurs the line between "action" and "test", unnecessarily. (The _caller_ is already testing that the language changed. _This_ code only needs to wait for the page to reload.)
2. This code appears to race: the user can open the old page's dropdown before the page reloads. After reload, subsequent lines will assume the dropdown is open (but it won't be).

I propose:

```
# wait for page reload
# 1. wait for the existing page to go away (by testing the modal is gone)
browser.assert_no_element(".modal.locale-switcher", wait=True)
# 2. wait for an element on the new page
browser.assert_element(".navbar .dropdown button", wait=True)

File: integrationtests/test_i18n.py [hunk 21]
```diff
@@ -2,38 +2,47 @@
...
     """
     browser.visit("/account/logout")
-    logout_button_text = {"en": "Log out", "el": "Αποσύνδεση"}
-    browser.click_button(logout_button_text[locale_id])
+    browser.click_whatever('.account_form button[type="submit"]')
+    browser.wait_for_element(".account_form.login")
+
+
+def switch_locale_django(browser: Browser, to_locale_name: str):
+    browser.click_whatever("a#locale-switcher-dropdown", wait=True)
```
adamhooper: Please nix `wait=True`, or add a long comment explaining what we're waiting for. (As I understand it, we don't need it. I've written more about `wait=True` in my next comment.)

adamhooper:
^^ This build failure came from master. Re-merging the fix from master to re-run the test.

PR #166, opened 2020-01-07
Use true ICU in python

giorgio93p:
This pull request replaces our custom quasi-ICU code with calls to ICU library, in order to support plurals, selects, and nested messages —all of which will be needed for module error messages.

### DIFF
--- a/cjworkbench/i18n/trans.py
+++ b/cjworkbench/i18n/trans.py
@@ -109,9 +109,9 @@ class MessageTranslator:
       - our HTML placeholder construct.
     """
 
-    def __init__(self, locale):
-        self.locale = locale
-        self.catalog = load_catalog(locale)
+    def __init__(self, locale_id):
+        self.icu_locale = Locale.createFromName(locale_id)
+        self.catalog = load_catalog(locale_id)
 
     def trans(self, message_id, default=None, context=None, parameters={}):
         """Find the message corresponding to the given ID in the catalog and format it according to the given parameters.
@@ -152,11 +152,11 @@ class MessageTranslator:
         if message:
             try:
                 return self._format_message(
-                    message, parameters=parameters, do_escape=False
+                    message, parameters=parameters, html_escape=False
                 )
             except (ICUError, InvalidICUParameters):
                 pass
-        return self._format_message(fallback, parameters=parameters, do_escape=False)
+        return self._format_message(fallback, parameters=parameters, html_escape=False)
 
     def _process_html_message(self, message, fallback, parameters={}, tags={}):
         if message:
@@ -164,12 +164,12 @@ class MessageTranslator:
                 return self._format_message(
                     self._replace_tags(message, tags),
                     parameters=parameters,
-                    do_escape=True,
+                    html_escape=True,
                 )
             except (ICUError, InvalidICUParameters):
                 pass
         return self._format_message(
-            self._replace_tags(fallback, tags), parameters=parameters, do_escape=True
+            self._replace_tags(fallback, tags), parameters=parameters, html_escape=True
         )
 
     def _replace_tags(self, target_message, tags):
@@ -180,7 +180,7 @@ class MessageTranslator:
         """
         return restore_tags(target_message, tag_mapping=tags)
 
-    def _format_message(self, message, parameters={}, do_escape=True):
+    def _format_message(self, message, parameters={}, html_escape=True):
         """Substitute parameters into ICU-style message.
         You can have variable substitution, plurals, selects and nested messages.
         
@@ -188,16 +188,14 @@ class MessageTranslator:
         """
         if not message:
             return message
-        message_format = MessageFormat(
-            UnicodeString(message), Locale.createFromName(self.locale)
-        )
+        message_format = MessageFormat(UnicodeString(message), self.icu_locale)
         if isinstance(parameters, dict):
             try:
                 return message_format.format(
                     [UnicodeString(x) for x in list(parameters.keys())],
                     [
                         Formattable(
-                            escape(x) if do_escape and isinstance(x, str) else x
+                            escape(x) if html_escape and isinstance(x, str) else x
                         )
                         for x in list(parameters.values())
                     ],
--- a/cjworkbench/i18n/trans.py
+++ b/cjworkbench/i18n/trans.py
@@ -20,7 +20,7 @@ def _get_translations(locale):
     """
     if locale in _translators:
         return _translators[locale]
-    _translators[locale] = MessageTranslator(locale)
+    _translators[locale] = MessageTranslator.for_application_messages(locale)
     return _translators[locale]
 
 
@@ -86,18 +86,22 @@ def restore_tags(message, tag_mapping):
 
 
 class MessageTranslator:
-    """Load the message catalogs for a given locale and provide helper methods for message translation.
+    """Hold a message catalog for a given locale and provide helper methods for message localization with ICU.
     It uses plaintext messages with variables in `{var}` placeholders, e.g. in `"Hello {name}"`, `name` is a variable.
     
     Essentially, it is a wrapper for the combination of 
-      - Babel (for loading PO files)
-      - variable substitution
+      - Babel (for reading messages of PO files)
+      - variable substitution with ICU
       - our HTML placeholder construct.
     """
 
-    def __init__(self, locale_id):
+    def __init__(self, locale_id, catalog):
         self.icu_locale = Locale.createFromName(locale_id)
-        self.catalog = load_catalog(locale_id)
+        self.catalog = catalog
+
+    @classmethod
+    def for_application_messages(cls, locale_id: str):
+        return cls(locale_id, load_catalog(locale_id))
 
     def trans(self, message_id, default=None, context=None, parameters={}):
         """Find the message corresponding to the given ID in the catalog and format it according to the given parameters.
--- a/cjworkbench/i18n/trans.py
+++ b/cjworkbench/i18n/trans.py
@@ -6,6 +6,14 @@ from cjworkbench.i18n import default_locale
 from cjworkbench.i18n.catalogs import load_catalog
 from string import Formatter
 from cjworkbench.i18n.exceptions import UnsupportedLocaleError, BadCatalogsError
+from icu import (
+    Formattable,
+    Locale,
+    MessageFormat,
+    UnicodeString,
+    ResourceBundle,
+    ICUError,
+)
 
 
 class InvalidICUParameters(Exception):
@@ -38,7 +46,24 @@ def trans(message_id, *, default, context=None, parameters={}):
     For code parsing reasons, respect the following order when passing keyword arguments:
         `message_id` and then `default` and then `context` and then everything else
     """
-    return _get_translations(get_language()).trans(
+    return do_trans(
+        get_language(),
+        message_id,
+        default=default,
+        context=context,
+        parameters=parameters,
+    )
+
+
+def do_trans(locale_id, message_id, *, default, context=None, parameters={}):
+    """Translate the given message ID to the given locale
+    
+    HTML is not escaped.
+    
+    For code parsing reasons, respect the following order when passing keyword arguments:
+        `message_id` and then `default` and then `context` and then everything else
+    """
+    return _get_translations(locale_id).trans(
         message_id, default=default, context=context, parameters=parameters
     )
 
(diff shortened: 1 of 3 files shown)


cjworkbench/i18n/trans.py
```diff
@@ -174,25 +199,25 @@ def _replace_tags(self, target_message, tags):
...
-        At this point, ICU is not actually supported (i.e. you can have no plurals, secects, etc).
-        Only variable substitution is supported.
+        You can have variable substitution, plurals, selects and nested messages.
         
         The parameters must be a dict
         """
         if not message:
             return message
+        message_format = MessageFormat(
+            UnicodeString(message), Locale.createFromName(self.locale)
```
adamhooper: I think now's the right time to change the `__init__()` function:

* Rename `locale` arg to `locale_id` and `self.locale` to `self.locale_id` should be `locale_id` (it's an ID, right?)
* `self.icu_locale = Locale.createFromName(locale_id)`

adamhooper: Can we nix these `UnicodeString`s? The code would be more legible without them. I don't know the edge cases, but I tested this:

```
>>> import icu
>>> mf = icu.MessageFormat("Hello, {person}!", icu.Locale.createFromName("en"))
>>> mf.format(["person"], [icu.Formattable("world")])
'Hello, world!'
```

cjworkbench/i18n/trans.py
```diff
@@ -174,25 +199,25 @@ def _replace_tags(self, target_message, tags):
 
     def _format_message(self, message, parameters={}, do_escape=True):
```
adamhooper: Should be `html_escape`, not `do_escape`.

cjworkbench/i18n/trans.py
```diff
@@ -174,25 +199,25 @@ def _replace_tags(self, target_message, tags):
...
-                            else parameters[key]
+                return message_format.format(
+                    [UnicodeString(x) for x in list(parameters.keys())],
+                    [
+                        Formattable(
+                            escape(x) if do_escape and isinstance(x, str) else x
                         )
-                        for key in parameters
-                    },
+                        for x in list(parameters.values())
```
adamhooper: Nix `list()`

cjworkbench/i18n/trans.py
```diff
@@ -174,25 +199,25 @@ def _replace_tags(self, target_message, tags):
...
             try:
-                return formatter.format(
-                    message,
-                    **{
-                        key: (
-                            escape(parameters[key])
-                            if do_escape and isinstance(parameters[key], str)
-                            else parameters[key]
+                return message_format.format(
+                    [UnicodeString(x) for x in list(parameters.keys())],
```
adamhooper: I would think this could simply be `list(parameters.keys())` -- no list comprehension.

cjworkbench/i18n/trans.py
```diff
@@ -174,25 +199,25 @@ def _replace_tags(self, target_message, tags):
...
+        You can have variable substitution, plurals, selects and nested messages.
         
         The parameters must be a dict
         """
         if not message:
             return message
+        message_format = MessageFormat(
+            UnicodeString(message), Locale.createFromName(self.locale)
+        )
         if isinstance(parameters, dict):
```
adamhooper: If `parameters` isn't a dict, that's a bug in our code. Please nix this if-statement so we crash like we should when there's an error.

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -56,6 +60,46 @@ def test_no_html_escape(self):
             'Hello <a href="/you?a=n&b=e">you > > there</a> my & friend',
         )
 
+    # Tests that plurals, selects and nested messages are fully supported
+    def test_icu_support(self):
+        message = (
+            "Hello {a}, you have {g, select,"
+            "   male {{n, plural,"
+            "       =0 {no boys} one {# little boy} other {# little boys}"
```
adamhooper: Could we please nix the word "little" from this test? It triggers me. (If I heard a random man on the street say something about "little boys," I'd suspect he's a pervert.)

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -72,23 +116,32 @@ def test_trans_params(self):
...
                 default="Hello {a} {0} {b}",
                 parameters={"a": "you", "0": "!", "b": "2"},
             ),
+            "Hello you ! 2",
+        )
 
     # Tests that a translator can't break our system by including a numeric variable in the message
     def test_format_invalid_message(self):
-        self.assertEqual(
+        self.assertTrue(
```
adamhooper: ... this should be a `self.assertEqual()`.

What is it equal to? I'm in suspense.

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -72,23 +116,32 @@ def test_trans_params(self):
...
             MessageTranslator(default_locale)._process_html_message(
                 "Hello {a} {0} {b}", "Hello {a} {b}", parameters={"a": "you", "b": "!"}
-            ),
-            "Hello you !",
+            )
+        )
+
+    # Tests that a translator can't break our system by adding or removing variables in the message
+    def test_format_translator_parameters(self):
+        self.assertTrue(
```
adamhooper: Again: should be a `self.assertEqual()` and I'm dying to know what it's equal to.

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -208,5 +262,51 @@ def test_trans_tag_placeholders(self):
...
-            '<span id="hi">Hello hello</span> &amp; <a href="/you">you</a> &lt; <a class="red big" href="/there?a=b&amp;c=d">there&lt;</a>!',
+            '<span id="hi">Hello hello</span> <span id="hi">&amp;</span> <a href="/you">you</a> &lt; <a class="red big" href="/there?a=b&amp;c=d">there&lt;</a>!',
+        )
+
+    # Tests that plurals, selects and nested messages are fully supported
+    def test_icu_support(self):
+        message = (
+            "Hello {a}, you have {g, select,"
+            "   male {{n, plural,"
+            "       =0 {<em0>no</em0> boys} one {# little boy} other {# little boys}"
```
adamhooper: Please nix "little" again :).

cjworkbench/tests/i18n/test_trans.py
adamhooper: Also: please test that we raise `icu.ICUError` when given an invalid message format.

```
>>> mf = icu.MessageFormat("Hello, {person person person}!", icu.Locale.createFromName("en"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
icu.ICUError: ('Syntax error in format pattern', 0, 8, 'Hello, {\x00\udc8c羘\x00藐\udc86羘\x00', 'person person p\x00'), error code: 65799
```

cjworkbench/i18n/trans.py
```diff
@@ -38,7 +46,24 @@ def trans(message_id, *, default, context=None, parameters={}):
...
+    return do_trans(
+        get_language(),
+        message_id,
+        default=default,
+        context=context,
+        parameters=parameters,
+    )
+
+
+def do_trans(locale_id, message_id, *, default, context=None, parameters={}):
```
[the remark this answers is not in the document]
giorgio93p: Indeed, I added it during experimenting with module serialization last month and I thought it will be useful in the future. However, it is not certain and it will not be hard to add it again if needed.

cjworkbench/i18n/trans.py
```diff
@@ -46,37 +32,20 @@ def trans(message_id, *, default, context=None, parameters={}):
...
-    """
-    return _get_translations(locale_id).trans(
+    return _get_translations(get_language()).trans(
         message_id, default=default, context=context, parameters=parameters
     )
 
 
-def trans_html(locale, message_id, *, default, context=None, parameters={}, tags={}):
-    """Translate the given message ID to the current locale
+def trans_html(locale_id, message_id, *, default, context=None, parameters={}, tags={}):
```
adamhooper: Thanks for simplifying! Now that the functions are right next to each other, we can see a consistency problem: `trans_html()` takes a `locale_id` parameter, but `trans()` doesn't. These two functions are beside each other, so they ought to have similar signatures. How can we make that happen?

cjworkbench/i18n/trans.py
```diff
@@ -169,24 +138,24 @@ def _process_simple_message(self, message, fallback, parameters={}):
         if message:
             try:
                 return self._format_message(
-                    message, parameters=parameters, do_escape=False
+                    message, parameters=parameters, html_escape=False
                 )
-            except (ICUError, InvalidICUParameters):
+            except Exception:
```
adamhooper: I much preferred the `(ICUError, InvalidICUParameters)`. Why did you switch back?

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -22,8 +23,26 @@ def test_trans_params(self):
...
+        self.assertEqual(
+            MessageTranslator(default_locale)._process_simple_message(
+                "Hello {a b}", "Hello {a} {b}", parameters={"a": "you", "b": "!"}
+            ),
+            "Hello you !",
+        )
+
     # Tests that a programmer can include a numeric variable in the message
-    def test_format_invalid_default(self):
+    def test_programmer_numeric_parameter(self):
```
adamhooper: Let's not test any distinction between "programmer" and "translator". `test_programmer_numeric_parameter()` and `test_translator_numeric_parameter()` do the exact same thing (though one calls a private method and the other doesn't).

_Every_ test involves both a programmer and a translator:

* `parameters` is set by a programmer.
* `message` is set by a translator
* `default` is set by ... who cares? (I hesitate to call the its author either "programmer" _or_ "translator".)

Don't test by "roles": test by _arguments_:

* Invalid `parameters` should raise an uncaught exception
* Invalid `message` should fallback to default (and we'll probably want to log the failure in the near future)
* Invalid `default` should raise an uncaught exception

cjworkbench/tests/i18n/test_trans.py
```diff
@@ -33,20 +52,24 @@ def test_format_invalid_default(self):
             "Hello you ! 2",
         )
 
-    # Tests that a translator can't break our system by including a numeric variable in the message
-    def test_format_invalid_message(self):
-        self.assertTrue(
+    # Tests that a translator can use a numeric variable in the message
```
adamhooper: Please test that an extra numeric variable in `message` doesn't break our system.

cjworkbench/i18n/trans.py
giorgio93p: I did prefer it too, but I realized that the only way I was creating `InvalidICUParameters` was a
```
except Exception:
   raise InvalidICUParameters
```
in the function that passes the parameters to ICU, without having any idea what `Exception` might be.

cjworkbench/i18n/trans.py
giorgio93p: I don't think we can make that happen. `trans_html` can be used in cases where the django locale is unset or wrong, for example when we send an email that data has been updated, hence it needs a locale parameter. On the other hand, `trans` is for cases where the locale of django can (or must) be trusted; currently, it is only used via `trans_lazy` to define placeholders in the signup form, which have to be locale-agnostic.

giorgio93p: We could remove `trans` altogether and only keep `trans_lazy`...

adamhooper: Thanks for explaining. I think long-term we'll want to make these signatures consistent; but seeing as it's difficult, I'm happy to merge as-is.

cjworkbench/i18n/trans.py
adamhooper: Oh, awesome! You nixed the exception altogether :).

The `except: pass` thing threw me for a loop. Here's what I think would be more natural:

At top of file:

```
import logging
...
logger = logging.getLogger(__name__)
```

and here:

```
try:
    return self._format_message(...)
except ICUError:
    logger.exception("Error in .po file")
    # ... and fall through to default
return self._format_message(...)
```

This strikes me as safer, because I'll be emailed at least once when there's an error :). "Crash + email Adam" is far, far better than "Never crash and never expose errors."

adamhooper: (I don't like to catch `Exception`. Let's enumerate the errors.)

giorgio93p: I added the logging in pull request #164, since it's already (and still) open, in order not to open a new pull request

PR #168, opened 2020-01-13
ASGI middleware for setting locale id in websocket requests

giorgio93p:


cjworkbench/middleware/i18n.py
```diff
@@ -33,7 +39,45 @@ def __call__(self, request):
...
 
-    def _decide_locale(self, request):
+
+class SetCurrentLocaleAsgiMiddleware:
+    def __init__(self, app):
+        self.app = app
+
+    def __call__(self, scope):
+        async def inner(receive, send):
+            # We modify the scope within `inner`,
```
adamhooper: This makes me worry. I'd rather do less by assuming more.

It makes me worry because there is no Django Channels middleware that adds keys to `scope` during the `inner()` call. Django Channels' own middleware uses a `LazyObject()` explicitly to avoid doing this. I don't know what would break; but since `channels.middleware.BaseMiddleware.__call__()` shallow-copies `scope`, I suspect modifying the scope within `inner` means the middleware can't compose with other middleware correctly. This could bite us later.

Here's where I'm coming from: Django Channels isn't adequately documented. I think of it as abandonware since the genius behind it has stepped away. Consider Django Channels _very_ fragile. Let's not push its limits.

To do less, we can assume that any incoming Websockets request was inspired by either 1) an HTML page served by Django; or 2) some sort of inane condition we don't care about.

Our Django middleware ensures a language cookie is set when the user logs in. So we don't need to use auth: if there's no language cookie, the user isn't logged in. (Right?)

What would break if we simply set `user=None` here during `LocaleDecider()` init?

And to extend the thought experiment: what would break if we were to nix the `user` argument from `LocaleDecider` _entirely_?

giorgio93p: Indeed, now that we have the cookie, we can omit user.

giorgio93p: And hence modify `scope` outside `inner`

PR #169, opened 2020-01-15
Unit test python/django extraction and nix support for context in trans

giorgio93p:
Support for message context in `trans` was interfering with message parameters. Since we don't use it (and are not planning to use it), I removed the support for it.

Now the behaviour of extraction scripts is unit tested. Also, `I18nMessage` type in cjwkernel has a trans function which is extracted normally.

`trans_html` was renamed to `localize_html`, in order to emphasize its different nature from `trans`; it's not expected to be parsed by pybabel and it expects a locale argument.

### DIFF
--- a/cjwkernel/types.py
+++ b/cjwkernel/types.py
@@ -621,14 +621,16 @@ class I18nMessage:
         message_id: str,
         *,
         default: str,
-        parameters: Dict[str, Union[int, float, str]] = {},
+        args: Dict[str, Union[int, float, str]] = {},
     ) -> I18nMessage:
         """
-        Build an I18nMessage.
+        Build an I18nMessage, marking it for translation.
+        
+        Use this method (instead of constructing `I18nMessage` directly) when you wish to mark a string for translation. Workbench's tooling will extract messages from all `I18nMessage.trans()` calls and send them to translators.
 
         The `default` argument is ignored at runtime, it's only used when parsing code.
         """
-        return cls(message_id, parameters)
+        return cls(message_id, args)
 
     @classmethod
     def from_dict(cls, value: Dict[str, Any]) -> I18nMessage:
--- a/cjwkernel/types.py
+++ b/cjwkernel/types.py
@@ -616,6 +616,21 @@ class I18nMessage:
         return cls("TODO_i18n", {"text": text})
 
     @classmethod
+    def trans(
+        cls,
+        message_id: str,
+        *,
+        default: str,
+        parameters: Dict[str, Union[int, float, str]] = {},
+    ) -> I18nMessage:
+        """
+        Build an I18nMessage.
+
+        The `default` argument is ignored at runtime, it's only used when parsing code.
+        """
+        return cls(message_id, parameters)
+
+    @classmethod
     def from_dict(cls, value: Dict[str, Any]) -> I18nMessage:
         return cls(value["id"], value["arguments"])
 
--- a/cjworkbench/i18n/catalogs/extract.py
+++ b/cjworkbench/i18n/catalogs/extract.py
@@ -122,7 +122,6 @@ def extract_python(
         # `messages` will have all the string parameters to our function
         # As we specify in the documentation of `trans`,
         # the first will be the message ID, the second will be the default message
-        # and the (optional) third will be the message context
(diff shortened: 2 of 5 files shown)


cjwkernel/types.py
```diff
@@ -615,6 +615,21 @@ def TODO_i18n(cls, text: str) -> I18nMessage:
...
+    def trans(
+        cls,
+        message_id: str,
+        *,
+        default: str,
+        parameters: Dict[str, Union[int, float, str]] = {},
+    ) -> I18nMessage:
+        """
+        Build an I18nMessage.
+
```
adamhooper: Please add a comment explaining why people should use this. My suggestion:

```
Use this method (instead of constructing `I18nMessage` directly) when you wish to mark a string for translation. Workbench's tooling will extract messages from all `I18nMessage.trans()` calls and send them to translators.
```

cjwkernel/types.py
```diff
@@ -615,6 +615,21 @@ def TODO_i18n(cls, text: str) -> I18nMessage:
         """
         return cls("TODO_i18n", {"text": text})
 
+    @classmethod
+    def trans(
+        cls,
+        message_id: str,
+        *,
+        default: str,
+        parameters: Dict[str, Union[int, float, str]] = {},
```
adamhooper: Let's call this `args`, so it corresponds with `I18nMesage.args`.

From briefly grepping, I surmise that the `parameters` argument in `cjworkbench.i18n.trans()` is ignored by the extraction logic. it _really_ could be called "args" or "arguments" everywhere. (ICU MessageFormat docs use the term, "arguments".) But I'm not compelled to change it everywhere, because then we'll have the same hassle with "args" here vs "arguments" elsewhere.

cjworkbench/i18n/catalogs/extract.py
```diff
@@ -119,35 +119,24 @@ def extract_python(
...
-            if len(messages) > 1 and messages[1]:
-                # If we have a default, add it as a special comment
-                # that will be processed by our `merge_catalogs` script
-                translator_comments.append(
-                    (message_lineno, "default-message: " + messages[1])
-                )
+        # `messages` will have all the string parameters to our function
+        # As we specify in the documentation of `trans`,
+        # the first will be the message ID, the second will be the default message
+        # and the (optional) third will be the message context
```
adamhooper: Nix the mention of "context" here.

cjworkbench/i18n/trans.py
```diff
@@ -28,38 +28,35 @@ def _get_translations(locale):
     return _translators[locale]
 
 
-def trans(message_id, *, default, context=None, parameters={}):
-    """Translate the given message ID to the current locale
-    
-    HTML is not escaped.
+def trans(message_id, *, default, parameters={}):
+    """Mark a message for translation and localize it to the current locale, not escaping HTML.
```
adamhooper: Let's nix "not escaping HTML" from this docstring. Mention of "HTML" makes sense when reading this file and when writing unit tests, but that's where its utility stops. For callers, I think it's more likely to mislead than to help.

(The mention of HTML implies that callers should _consider_ HTML-escaping, more than they would consider this with other string functions. But you don't see "not escaping HTML" in the `str.join()` docstring, for instance. And even if we started using `trans()` within HTML templates, one could simply write `{% trans xxx, xxx, xxx %}` without thinking of HTML escaping, just like with any other function.

cjworkbench/i18n/trans.py
```diff
@@ -28,38 +28,35 @@ def _get_translations(locale):
...
-def trans_html(locale_id, message_id, *, default, context=None, parameters={}, tags={}):
-    """Translate the given message ID to the given locale
+trans_lazy = lazy(trans)
+"""Mark a string for translation, but actually localize it when it has to be used.
+   See the documentation of `trans` for more details on the function and its parameters.
+"""
+
+
+def localize_html(
+    locale_id, message_id, *, default, context=None, parameters={}, tags={}
```
adamhooper: Is it right that we still have `context` here?

giorgio93p: It is. Due to different parsing algorithm, we can support context in template tags. I also test for it in the new extraction unit tests.

PR #170, opened 2020-01-17
Add source field to I18nMessage

giorgio93p:


cjwkernel/pandas/types.py
```diff
@@ -411,12 +411,19 @@ class I18nMessage:
     args: Dict[str, Union[int, float, str]] = field(default_factory=dict)
     """Arguments (empty if message does not need any -- which is common)."""
 
+    source: Dict[str, str] = field(default_factory=dict)
```
adamhooper: I think this should be `Optional[Dict[...]] = None`, to be consistent with my suggestions in `moduletypes.py` and `cjwkernel/types.py`.

cjwkernel/thrift.thrift
```diff
@@ -203,6 +203,14 @@ struct I18nMessage {
    * For instance, `{"nColumns": 3, "exampleColumn": "Column X"}`
    */
   2: map<string, I18nArgument> arguments
+  
+  /**
+   * An indication of where the message is coming from.
+   * - An empty map means it's coming from workbench itself
+   * - A map with key `"module_id"` means it's coming from a module
```
adamhooper: Thrift is strongly-typed. This should be a `3: I18nMessageSource source`: a union of a `module_id` str or a `library` str. See `ParamValue` for an example of declaring union values. (Basically, a union looks remarkably similar to a struct.)

The comment in `ParamValue` also describes how we can achieve `None`.

cjwkernel/types.py
```diff
@@ -592,16 +592,26 @@ class I18nMessage:
     args: Dict[str, Union[int, float, str]] = field(default_factory=dict)
     """Arguments (empty if message does not need any -- which is common)."""
 
+    source: Dict[str, str] = field(default_factory=dict)
```
adamhooper: I was envisioning:

```
@dataclass(frozen=True)
class I18nMessageSource:
    pass

@dataclass(frozen=True)
class I18nMessageSourceLibrary(I18nMessageSource):
    library: str

@dataclass(frozen=True)
class I18nMessageSourceModule(I18nMessageSource):
    module_id: str

I18nMessageSource.Library = I18nMessageSourceLibrary
I18nMessageSource.Module = I18nMessageSourceModule

# ... and then here, within I18nMessage:
    source: Optional[I18nMessageSource]
```

My reasoning behind `Optional`: our database is already chock-full of `I18nMessage` values (all `{"id": "TODO_i18n", {"text": "..."}}`), and in _those_ ones, `source` is undefined. So I'd feel uncomfortable storing `source: {}` in other `I18nMessage` values in the database, because then we'd have two representations of the same thing in the database.

My reasoning behind strongly-typed data classes: that's what this module is all about. (See `ColumnType`, for instance.)

cjwkernel/types.py
```diff
@@ -634,10 +644,10 @@ def trans(
 
     @classmethod
     def from_dict(cls, value: Dict[str, Any]) -> I18nMessage:
-        return cls(value["id"], value["arguments"])
+        return cls(value["id"], value["arguments"], value["source"])
```
adamhooper: This isn't backwards-compatible. We already have thousands of I18nMessages in our database, and none of them have a `source`. Use `value.get("source", None)`.

cjwkernel/types.py
```diff
@@ -634,10 +644,10 @@ def trans(
...
 
     def to_dict(self) -> Dict[str, Any]:
-        return {"id": self.id, "arguments": self.args}
+        return {"id": self.id, "arguments": self.args, "source": self.source}
```
adamhooper: To stay consistent with the thousands of I18nMessages already in our database, don't produce a `source` key if `self.source is None`.

cjwkernel/pandas/types.py
```diff
@@ -411,12 +411,19 @@ class I18nMessage:
...
+    """An indication of where the message is coming from.
+        - An empty dict means it's coming from workbench itself
+        - A dict with key `"module_id"` means it's coming from a module
```
adamhooper: `"module"`

cjwkernel/pandas/types.py
```diff
@@ -411,19 +411,41 @@ class I18nMessage:
...
     @classmethod
     def from_arrow(cls, value: atypes.I18nMessage) -> I18nMessage:
-        return cls(value.id, value.args, value.source)
+        if not value.source:
+            return cls(value.id, value.args)
+        if isinstance(value.source, atypes.I18nMessageSource.Module):
+            return cls(value.id, value.args, {"module": value.source.module_id})
+        if isinstance(value.source, atypes.I18nMessageSource.Library):
+            return cls(value.id, value.args, {"library": value.source.library})
+        raise ValueError("Unknown I18nMessage source")
```
adamhooper: Don't tempt callers to recover from undefined behavior.

`ValueError` suggests to me that somewhere, somehow, a _user_ entered the wrong data. And it suggests that a caller may want to catch the error.

But if this exception is raised, _cjworkbench itself_ is buggy. So if there _must_ be an exception, make it one that nobody would _dare_ catch -- for instance, `assert False` (AssertionError) or `raise NotImplementedError` (a RuntimeError).

In this specific function, there's a better approach: rearrange the if-statements so the `return cls(value.id, value.args)` comes in an `else:` clause.

Try to picture an integration-test suite that leads to 100% code coverage. It will never hit this `raise` line. So the line shouldn't exist.

cjwkernel/pandas/types.py
```diff
@@ -411,19 +411,41 @@ class I18nMessage:
...
+                self.args,
+                atypes.I18nMessageSource.Module(self.source["module"]),
+            )
+        if "library" in self.source:
+            return atypes.I18nMessage(
+                self.id,
+                self.args,
+                atypes.I18nMessageSource.Library(self.source["library"]),
+            )
+        raise ValueError("Unknown I18nMessage source")
```
adamhooper: Again, avoid raising an error that can never occur.

cjwkernel/pandas/types.py
```diff
@@ -469,15 +491,15 @@ def coerce(cls, value: mtypes.Message) -> I18nMessage:
...
                 if len(source) == 1 and set(source.keys()) - supported_source_keys:
                     raise ValueError(
                         "Unknown message source kind %s. Supported kinds: %s."
                         % (source.keys(), supported_source_keys)
                     )
             else:
-                source = {}
-            return cls(value[0], value[1], source)
+                source = None
+            return cls(value[0], value[1], source or None)
```
adamhooper: I don't understand the `source or None`.

Please assert above that `len(source) == 1`. And then I think you can nix the `or None` here, right? (Otherwise, please add a comment saying what it does.)

cjwkernel/thrift.thrift
```diff
@@ -192,6 +192,22 @@ union I18nArgument {
   3: double double_value
 }
 
+/** Source of a translatable string. 
+    If none of the values are set, this means it's coming from workbench itself
```
adamhooper: Please follow to the docstring convention from the rest of this file. (By my count, this comment needs two more newlines and three more asterisks.)

cjwkernel/thrift.thrift
```diff
@@ -192,6 +192,22 @@ union I18nArgument {
...
+*/
+union I18nMessageSource {
+  /**
+   * If `"library"` is set, this means it's coming from some of our supported libraries
```
adamhooper: Nix the `"If ... is set, this means it's coming from"` in these two comments. (The reader must read them all to determine they can all be ignored.) Instead:

```
/** The message comes from the "library" library */
...
/** The message comes from the "module_id" module */
```

cjwkernel/thrift.thrift
```diff
@@ -206,11 +222,8 @@ struct I18nMessage {
   
   /**
    * An indication of where the message is coming from.
```
adamhooper: `"Where the message comes from."`

cjwkernel/types.py
```diff
@@ -582,6 +582,67 @@ def _i18n_argument_to_thrift(value: Union[str, int, float]) -> ttypes.I18nArgume
...
+        cls, value: ttypes.I18nMessageSource
+    ) -> Optional[I18nMessageSource]:
+        if value.module_id is not None:
+            return I18nMessageSource.Module(value.module_id)
+        elif value.library is not None:
+            return I18nMessageSource.Library(value.library)
+        else:
+            return None
+
+    @abstractmethod
```
adamhooper: This is not an abstract method. Nix `@abstractmethod` and add an `# override` comment instead.

cjwkernel/types.py
```diff
@@ -582,6 +582,67 @@ def _i18n_argument_to_thrift(value: Union[str, int, float]) -> ttypes.I18nArgume
...
+    @classmethod
+    def from_dict(cls, value: Dict[str, Any]) -> Optional[I18nMessageSource]:
+        if value.get("module") is not None:
+            return I18nMessageSource.Module(value["module"])
+        elif value.get("library") is not None:
+            return I18nMessageSource.Library(value["library"])
+        else:
+            return None
+
+    @abstractmethod
```
adamhooper: Not an abstract method.

cjwkernel/types.py
```diff
@@ -592,26 +653,22 @@ class I18nMessage:
...
     """Arguments (empty if message does not need any -- which is common)."""
 
-    source: Dict[str, str] = field(default_factory=dict)
-    """An indication of where the message is coming from.
-        - An empty dict means it's coming from workbench itself
-        - A dict with key `module_id` means it's coming from a module
-        - A dict with key `library` indicates it's coming from some of our supported libraries (e.g. `cjwmodule`)
-    """
+    source: Optional[I18nMessageSource] = None
+    """An indication of where the message is coming from. `None` means it's coming from workbench itself."""
```
adamhooper: `"""Where the message comes from, or `None` if it comes from Workbench proper."""`

(trying to fit 80 char line limit)

PR #173, opened 2020-01-20
Remove a signal receiver that creates problems

giorgio93p:
When `allauth` saved a new user during social login, the deleted receiver would create a default user profile (hence with the default locale id). That default profile would be used later, instead of the updated one, when trying to read the user profile and write the cookie locale, resulting in the user seeing the default locale after login and not the locale in their (updated) profile.

### DIFF
--- a/server/tests/handlers/test_wf_module.py
+++ b/server/tests/handlers/test_wf_module.py
@@ -565,9 +565,7 @@ class WfModuleTest(HandlerTestCase):
 
     def test_try_set_autofetch_exceed_quota(self):
         user = User.objects.create(username="a", email="a@example.org")
-        UserProfile.objects.update_or_create(
-            user=user, defaults={"max_fetches_per_day": 10}
-        )
+        UserProfile.objects.create(user=user, max_fetches_per_day=10)
         workflow = Workflow.create_and_init(owner=user)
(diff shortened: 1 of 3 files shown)


server/tests/handlers/test_wf_module.py
```diff
@@ -564,8 +565,9 @@ def test_try_set_autofetch_disable_autofetch(self):
 
     def test_try_set_autofetch_exceed_quota(self):
         user = User.objects.create(username="a", email="a@example.org")
-        user.user_profile.max_fetches_per_day = 10
-        user.user_profile.save()
+        UserProfile.objects.update_or_create(
```
adamhooper: When does this update an existing object?

If never, use `.create()`.

server/tests/handlers/test_wf_module.py
```diff
@@ -588,8 +590,9 @@ def test_try_set_autofetch_exceed_quota(self):
 
     def test_try_set_autofetch_allow_exceed_quota_when_reducing(self):
         user = User.objects.create(username="a", email="a@example.org")
-        user.user_profile.max_fetches_per_day = 10
-        user.user_profile.save()
+        UserProfile.objects.update_or_create(
```
adamhooper: `.create()`?

PR #174, opened 2020-01-20
Internationalize error messages in renderer.execute.types

giorgio93p:
This pull request replaces English `TODO_i18n` error texts in `renderer.execute.types` with ICU `I18nMessage`s. In order for them to be able to get localized, the pull request also makes use of recent changes in asgi middleware, serialization, and `I18nMessage`s.

### DIFF
--- a/cjworkbench/i18n/trans.py
+++ b/cjworkbench/i18n/trans.py
@@ -31,10 +31,10 @@ _TagMapping = Dict[str, _Tag]
 """ Maps each tag to its data
 """
 
-_Parameters = Dict[str, Union[int, float, str]]
+_MessageArguments = Dict[str, Union[int, float, str]]
 
 
-def trans(message_id: str, *, default: str, parameters: _Parameters = {}) -> str:
+def trans(message_id: str, *, default: str, arguments: _MessageArguments = {}) -> str:
     """Mark a message for translation and localize it to the current locale.
 
     `default` is only considered when parsing code for message extraction.
@@ -44,16 +44,16 @@ def trans(message_id: str, *, default: str, parameters: _Parameters = {}) -> str
     For code parsing reasons, respect the following order when passing keyword arguments:
         `message_id` and then `default` and then everything else
     """
-    return localize(get_language(), message_id, parameters=parameters)
+    return localize(get_language(), message_id, arguments=arguments)
 
 
 trans_lazy = lazy(trans)
 """Mark a string for translation, but actually localize it when it has to be used.
-   See the documentation of `trans` for more details on the function and its parameters.
+   See the documentation of `trans` for more details on the function and its arguments.
 """
 
 
-def localize(locale_id: str, message_id: str, parameters: _Parameters = {}) -> str:
+def localize(locale_id: str, message_id: str, arguments: _MessageArguments = {}) -> str:
     """Localize the given message ID to the given locale.
 
     Raise `KeyError` if the message is not found (neither in the catalogs of the given and of the default locale).
@@ -63,7 +63,7 @@ def localize(locale_id: str, message_id: str, parameters: _Parameters = {}) -> s
     if locale_id != default_locale:
         try:
             return _get_translations(locale_id).get_and_format_message(
-                message_id, parameters=parameters
+                message_id, arguments=arguments
             )
         except ICUError as err:
             logger.exception(
@@ -72,7 +72,7 @@ def localize(locale_id: str, message_id: str, parameters: _Parameters = {}) -> s
         except KeyError as err:
             pass
     return _get_translations(default_locale).get_and_format_message(
-        message_id, parameters=parameters
+        message_id, arguments=arguments
     )
 
 
@@ -80,7 +80,7 @@ def localize_html(
     locale_id: str,
     message_id: str,
     context: Optional[str] = None,
-    parameters: _Parameters = {},
+    arguments: _MessageArguments = {},
     tags: _TagMapping = {},
 ) -> str:
     """Localize the given message ID to the given locale, escaping HTML.
@@ -88,12 +88,12 @@ def localize_html(
     Raise `KeyError` if the message is not found (neither in the catalogs of the given and of the default locale).
     Raise `ICUError` if the message in the default locale is incorrectly formatted.
     
-    HTML is escaped in the message, as well as in parameters and tag attributes.
+    HTML is escaped in the message, as well as in arguments and tag attributes.
     """
     if locale_id != default_locale:
         try:
             return _get_translations(locale_id).get_and_format_html_message(
-                message_id, context=context, parameters=parameters, tags=tags
+                message_id, context=context, arguments=arguments, tags=tags
             )
         except ICUError as err:
             logger.exception(
@@ -102,7 +102,7 @@ def localize_html(
         except KeyError as err:
             pass
     return _get_translations(default_locale).get_and_format_html_message(
-        message_id, context=context, parameters=parameters, tags=tags
+        message_id, context=context, arguments=arguments, tags=tags
     )
 
 
@@ -157,47 +157,47 @@ class MessageTranslator:
         self,
         message_id: str,
         context: Optional[str] = None,
-        parameters: _Parameters = {},
+        arguments: _MessageArguments = {},
     ) -> str:
-        """Find the message corresponding to the given ID in the catalog and format it according to the given parameters.
+        """Find the message corresponding to the given ID in the catalog and format it according to the given arguments.
         If the message is either not found or empty, a `KeyError` is raised.
         If the message is incorrectly formatted, an `ICUError` is raised.
         
-        See `self._format_message` for acceptable types of the `parameters` argument.
+        See `self._format_message` for acceptable types of the `arguments` argument.
         """
         return self._process_simple_message(
-            self.get_message(message_id, context=context), parameters
+            self.get_message(message_id, context=context), arguments
         )
 
     def get_and_format_html_message(
         self,
         message_id: str,
         context: Optional[str] = None,
-        parameters: _Parameters = {},
+        arguments: _MessageArguments = {},
         tags: _TagMapping = {},
     ) -> str:
-        """Find the message corresponding to the given ID in the catalog and format it according to the given parameters.
+        """Find the message corresponding to the given ID in the catalog and format it according to the given arguments.
         If the message is either not found or empty, a `KeyError` is raised.
         If the message is incorrectly formatted, an `ICUError` is raised.
         
-        See `self._format_message` for acceptable types of the parameters argument.
+        See `self._format_message` for acceptable types of the arguments argument.
         
         HTML-like tags in the message used are replaced by their counterpart in `tags`, as specified in `restore_tags`
         """
         return self._process_html_message(
-            self.get_message(message_id, context=context), parameters, tags
+            self.get_message(message_id, context=context), arguments, tags
         )
 
     def _process_simple_message(
-        self, message: str, parameters: _Parameters = {}
+        self, message: str, arguments: _MessageArguments = {}
     ) -> str:
-        return self._format_message(message, parameters=parameters, html_escape=False)
+        return self._format_message(message, arguments=arguments, html_escape=False)
 
     def _process_html_message(
-        self, message: str, parameters: _Parameters = {}, tags: _TagMapping = {}
+        self, message: str, arguments: _MessageArguments = {}, tags: _TagMapping = {}
     ) -> str:
         return self._format_message(
-            self._replace_tags(message, tags), parameters=parameters, html_escape=True
+            self._replace_tags(message, tags), arguments=arguments, html_escape=True
         )
 
     def _replace_tags(self, target_message: str, tags: _TagMapping) -> str:
@@ -209,20 +209,20 @@ class MessageTranslator:
         return restore_tags(target_message, tag_mapping=tags)
 
     def _format_message(
-        self, message: str, parameters: _Parameters = {}, html_escape: bool = True
+        self, message: str, arguments: _MessageArguments = {}, html_escape: bool = True
     ) -> str:
-        """Substitute parameters into ICU-style message.
+        """Substitute arguments into ICU-style message.
         You can have variable substitution, plurals, selects and nested messages.
         
         Raises `ICUError` in case of incorrectly formatted message.
         
-        The parameters must be a dict
+        The arguments must be a dict
         """
         return MessageFormat(message, self.icu_locale).format(
-            list(parameters.keys()),
+            list(arguments.keys()),
             [
                 Formattable(escape(x) if html_escape and isinstance(x, str) else x)
-                for x in parameters.values()
+                for x in arguments.values()
             ],
         )
 
--- a/cjworkbench/i18n/trans.py
+++ b/cjworkbench/i18n/trans.py
@@ -33,13 +33,12 @@ _TagMapping = Dict[str, _Tag]
 _Parameters = Dict[str, Union[int, float, str]]
 
 
-def trans(
-    message_id: str, *, default: str, parameters: _Parameters = {}
-) -> Optional[str]:
+def trans(message_id: str, *, default: str, parameters: _Parameters = {}) -> str:
     """Mark a message for translation and localize it to the current locale.
 
     `default` is only considered when parsing code for message extraction.
-    If the message is not found in the catalog for the current or the default locale, return `None`.
+    If the message is not found in the catalog for the current or the default locale, return `None`,
+    raise `KeyError`.
     
     For code parsing reasons, respect the following order when passing keyword arguments:
         `message_id` and then `default` and then everything else
@@ -53,16 +52,19 @@ trans_lazy = lazy(trans)
 """
 
 
-def localize(
-    locale_id: str, message_id: str, parameters: _Parameters = {}
-) -> Optional[str]:
+def localize(locale_id: str, message_id: str, parameters: _Parameters = {}) -> str:
     """Localize the given message ID to the given locale.
 
-    If the message is not found in the catalog for the given or the default locale, return `None`.
+    If the message is not found in the catalog for the given or the default locale, return `None`,
+    raise `KeyError`.
     """
-    return _get_translations(locale_id).trans(
+    message = _get_translations(locale_id).trans(
         message_id, parameters=parameters
     ) or _get_translations(default_locale).trans(message_id, parameters=parameters)
+    if message:
+        return message
+    else:
+        raise KeyError(message_id)
 
 
 def localize_html(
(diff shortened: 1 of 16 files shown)


renderer/execute/types.py
```diff
@@ -128,31 +130,38 @@ def _as_i18n_message(self) -> I18nMessage:
...
+                return I18nMessage.trans(
+                    "py.renderer.execute.types.PromptingError.WrongColumnType.as_error_message.general",
+                    default="{ columns, plural, offset:2"
+                    " =1 {The column “{0}” must be converted from { found_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}} to {best_wanted_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}}.}"
+                    " =2 {The columns “{0}” and “{1}” must be converted from { found_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}} to {best_wanted_type, select, text {Text} number {Numbers} datetime {Dates & Times}  other{}}.}"
+                    " one {The columns “{0}”, “{1}” and “{2}” must be converted from { found_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}} to {best_wanted_type, select, text {Text} number {Numbers} datetime {Dates & Times} other{}}.}"
```
adamhooper: I'd have put `=3` here. I had to read the docs on the `offset` param to figure out why `one` selects `3`.

server/serializers.py
```diff
@@ -265,10 +267,15 @@ def jsonize_clientside_tab(tab: clientside.TabUpdate) -> Dict[str, Any]:
 
 
 def jsonize_i18n_message(message: I18nMessage, ctx: JsonizeContext) -> str:
+    assert message.source is None
     if message.id == "TODO_i18n":
         return message.args["text"]
     else:
-        raise RuntimeError("TODO_i18n")
+        return localize(
```
adamhooper: What triggers the `or`? I think we need a comment here.

I feel bad asking you to do something I myself neglected to do, but ... well, I will :). Could you please create `server/tests/test_serializers.py` and unit-test the three code paths here? (`TODO_i18n`; `localize(ctx.locale_id, ...)`; `not localize(ctx.locale_id, ...)`)

renderer/execute/types.py
```diff
@@ -128,31 +130,38 @@ def _as_i18n_message(self) -> I18nMessage:
...
-                return I18nMessage.TODO_i18n(
-                    f"{columns_str} must be converted "
-                    f"from {self.found_type_name} to {self.best_wanted_type_name}."
+                icu_args["found_type"] = self.found_type
+                icu_args["best_wanted_type"] = self.best_wanted_type_id
+                # i18n: The parameter {columns} will contain the total number of columns that need to be converted; you will also receive the column names: {0}, {1}, {2}, etc. The parameters {found_type} and {best_wanted_type} will have values among "text", "number", "datetime"; however, including a (possibly empty) "other" case is mandatory.
+                return I18nMessage.trans(
+                    "py.renderer.execute.types.PromptingError.WrongColumnType.as_error_message.general",
+                    default="{ columns, plural, offset:2"
+                    " =1 {The column “{0}” must be converted from { found_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}} to {best_wanted_type, select, text {Text} number {Numbers} datetime {Dates & Times} other {}}.}"
```
adamhooper: We plan to add new types to Workbench. And I've _almost_ convinced Pierre that we should rename "Dates & Times" to "Timestamps" (making way for a new "Date" type). I worry it will be costly to maintain these `select` clauses in all pofiles.

Should we translate the types separately?

giorgio93p: In general, it's difficult (or impossible) for the result to make (ideal) sense in any possible (human) language when we translate it in parts and combine them with code. Even if we assume that the syntax will make sense (as we are tempted -and, who knows, maybe entitled?- to do in this case), some languages are heavily inflected (i.e. may need different forms for some terms depending on their syntactic function); for example, a simple Greek translation of the message you commented on would require the `best_wanted_type` translations to be in accusative, but requiring a translation in accusative might conflict with other languages.

Another reason that we can't compose messages is the current structure of `I18nMessage`s, although we could overcome this. Currently, the parameters of an `I18nMessage` can only be plain data (number or string); if we want to compose messages, we must allow the type `I18nMessage` for parameters of `I18nMessage`s.

renderer/execute/types.py
giorgio93p: Indeed, `offset` is kind of peculiar; I had to experiment to end up with the correct cases. However, it's the recommended way to make messages that read like `A and B and 4 others`.

server/serializers.py
giorgio93p: The intention is this procedure
- localize in given locale, if there is a message for it
- localize in default locale, if there is a message for it
- return message id, which is better than nothing; we could also throw an exception here if you like

I will add a comment and unit tests

renderer/execute/types.py
adamhooper: We don't inflect column names; and I don't think users expect us to. Could some creative phrasing, punctuation or formatting convey "this word is data, not a normal word?" Quotation marks / small-caps / underline / ...?

As for the structure: you make a good point -- there's a solid challenge here. I can imagine customizing ICU to let Greek translators write `{found_type, cjwcolumntype, accusative}`. ICU does not [encourage](http://userguide.icu-project.org/formatparse/messages#TOC-Custom-Format-Objects-discouraged-) this, but it's probably doable. This would involve some careful code per language, once.

So we're on the same page, could you please describe to me what we'd need to do to add a new type to Workbench using the strategy of this pull request as-is? Keep in mind that modules' messages may include data-type names, too. And when the number of Workbench modules explodes, we on the core team won't be able to revise modules that aren't ours.

My thesis: we're committing to a _ton_ of effort here, regardless of our choice.

renderer/execute/types.py
adamhooper: What I mean is: I think the `one` should be `=3`. That way the `offset:2` only affects the behavior of `#`, which seems more intuitive.

server/serializers.py
adamhooper: Ohhh, I had misunderstood the `localize()` function's behavior. Let's take our time here....

First: could you please tweak the docs in `i18n/trans.py` to clarify this? The `localize()` method should:

* Be type-annotated. It returns `Optional[str]` -- that's not obvious at all.
* Explicitly state: `Return None if there is no translation in the given locale and no default.`
* While we're at it, explicitly state: `Raise ICUError if the default message is invalid (programmer error).`

Ditto for `trans()`.

My instinct tells me we're going about this the wrong way here -- this double-invocation of `localize()` feels different from our `trans()` calls. We could make them feel the same by making `MessageTranslator` use the default _catalog_ instead of the default _message_. (`MessageTranslator.trans()` would not take a `default` argument.) Does that sound tempting?

adamhooper: (Or another idea: `localize()` could look up `default` itself from the default catalog and pass it to `MessageTranslator.trans()`. That might give the same outcome with fewer code changes.)

giorgio93p: The ICU locale stored in `MessageTranslator` should match the locale of the catalog stored in `MessageTranslator`. If it does not, there will be inconsistent behaviour; for example when asking for a Ukranian text ([they use](http://cldr.unicode.org/index/cldr-spec/plural-rules) the `one` rule also for some values ending in 1) and an English message comes up from the default catalog, the case for `one` will be used when the value is 21, 31, etc. Moreover, `MessageTranslator` already has many responsibilities (fetch message from catalog, replace parameters, and replace tags if needed).

We could move the multiple invocations within `trans`, `localize`, and `localize_html`: they could get the `MessageTranslator` for their given locale, fall back to the `MessageTranslator` for the default locale, and then fall back to `default`; they could even ignore (`trans`) or not have (`localize` and `localize_html`) the `default` argument. Without `default`, they could return `None` or the `message_id`.

adamhooper: Your concern about `MessageTranslator` makes me realize ... we already have a bug here, right? When `MessageTranslator.trans()` translates the fallback, it does it in the catalog locale, _not_ English.

It would certainly feel comfortable to me if `localize()` didn't have a `default` argument -- and if the `trans()` implementation simply called `localize()`. If you like that approach, too, please go for it.

All of this suggests to me a direction that would clarify the `localize()` API and simplify `MessageTranslator` a bit:

* Nix `default` from `MessageTranslator.trans()`, and move the fallback logic into `localize()` itself.
* Change the module-level `trans()` to invoke `localize()`

For this pull request, I'm only _really_ concerned with the `localize()` _API_. (These fallbacks are a low priority since they're never meant to show up.) But I'm curious if you agree that A) the ICU-locale bug you describe already exists; and B) my suggestion would address it neatly.

giorgio93p: Oops! Indeed, this bug already exists in `MessageTranslator`! Your suggestion looks great.

renderer/execute/types.py
giorgio93p: I wasn't referring to inflected column names, but to inflected column types. If we translate data types separately, some trick such as the `column "A"` we use for the column name could probably be used by the translators for the column types (I don't know if it would sound ideal in _all_ languages, but it would probably work for all languages in some way). But, of course, we would need to make `I18nMessage` able to store `I18nMessage` as a parameter.

With the current status of the pull request, in order to add a new data type, we would indeed need to add a new case to all the `select` messages (anywhere the po files are stored) that have data types as cases.

`{found_type, cjwcolumntype, accusative}` looks kind of weird

adamhooper: Thank you for talking all this through with me. I'm inclined to table my concern for now. We can redesign the types approach later, if we find the `select` clauses are too costly in practice.

server/serializers.py
giorgio93p: I refactored and added a few more unit tests. Now `trans` and `jsonize_i18n_message` use `localize`, while `trans_html` (template tag) uses `localize_html`.  `localize` and `localize_html` do the "try this locale, else try default locale, else return None" stuff and `MessageTranslator` is not concerned with defaults at all. It looks way better.

I opted for everything returning `None` when no messages have been found. This will prompt developers to run `bin/dev extract-messages` earlier, since they will not see their texts otherwise. Keep in mind that it would also make some sense to make the upper-level functions, i.e. `trans`, `trans_html` (template tag) and `jsonize_18n_message` return the message id instead of `None`; we can decide it later.

cjworkbench/i18n/trans.py
```diff
@@ -45,26 +53,39 @@ def trans(message_id, *, default, parameters={}):
...
 
 
-def localize(locale_id, message_id, *, default, parameters={}):
-    """Localize the given message ID to the given locale."""
+def localize(
+    locale_id: str, message_id: str, parameters: _Parameters = {}
+) -> Optional[str]:
+    """Localize the given message ID to the given locale.
+
+    If the message is not found in the catalog for the given or the default locale, return `None`.
```
adamhooper: This forces _all_ callers of `localize()` and `trans()` to use an if-statement with the return value. `None` is not interchangeable with `str`. For instance, if we serialize a module param name of `null`, then our JS side will crash after printing a slew of proptype errors. And our test suite won't catch the crash.

Please `raise KeyError(module_id)` if the message does not exist. Callers that actually _want_ special behavior when the key doesn't exist may catch the exception.

cjworkbench/i18n/trans.py
```diff
@@ -197,13 +227,23 @@ def _format_message(self, message, parameters={}, html_escape=True):
...
         """
-        if context:
-            message = self.catalog.get(message_id, context)
-        else:
-            message = self.catalog.get(message_id)
-        return message.string if message else None
+        return find_string(self.catalog, message_id, context=context)
+
+
+def _get_translations(locale: str) -> MessageTranslator:
```
adamhooper: `locale_id`

cjworkbench/i18n/trans.py
```diff
@@ -197,13 +227,23 @@ def _format_message(self, message, parameters={}, html_escape=True):
...
+    """Return a singleton MessageTranslator for the given locale.
+    
+    In order to parse the message catalogs only once per locale,
+    uses the _translators dict to store the created MessageTranslator for each locale.
+    """
+    if locale in _translators:
+        return _translators[locale]
+    _translators[locale] = MessageTranslator.for_application_messages(locale)
```
adamhooper: This looks like it should be `_get_translations = lru_cache()(MessageTranslator.for_application_messages)`. Should it? If not, please comment explaining why.

(You can keep the docstring by placing it on the line after the assignment.)

server/serializers.py
```diff
@@ -266,7 +266,13 @@ def jsonize_clientside_tab(tab: clientside.TabUpdate) -> Dict[str, Any]:
     return d
 
 
-def jsonize_i18n_message(message: I18nMessage, ctx: JsonizeContext) -> Optional[str]:
+def jsonize_i18n_message(message: I18nMessage, ctx: JsonizeContext) -> str:
+    """Localize (or unwrap, if it's a TODO_i18n) an `I18nMessage`
+    
+    Uses `locale_id` from `ctx`
+    
+    Raises `KeyError` if the message text cannot be found in the catalogs.
```
adamhooper: In this particular case -- which is not general -- the `I18nMessage` comes from a _module_, not from Workbench. A bug in a module should not make Workbench crash.

How about: "If the message text is not in the user-locale catalog or the default-locale catalog, log a KeyError and return a JSON-serialized representation of the I18nMessage."

And: "If the message text in the default locale cannot be formatted correctly, log an ICUError and return a JSON-serialized representation of the I18nMessage."

cjworkbench/i18n/trans.py
```diff
@@ -53,16 +53,19 @@ def trans(
...
-) -> Optional[str]:
+def localize(locale_id: str, message_id: str, parameters: _Parameters = {}) -> str:
     """Localize the given message ID to the given locale.
 
-    If the message is not found in the catalog for the given or the default locale, return `None`.
+    If the message is not found in the catalog for the given or the default locale, return `None`,
+    raise `KeyError`.
     """
-    return _get_translations(locale_id).trans(
+    message = _get_translations(locale_id).trans(
```
adamhooper: This changes the behavior we had before. I feel the previous behavior was better. Thoughts?

Previously, `trans()` would:

1. Try the user-locale message catalog. If there's an ICUError, log the error and fall through.
2. Try the default-locale message catalog. If there's an ICUError, raise it.

After applying this pull request, we get:

1. Try the user-locale message catalog. If there's an ICUError, log the error and fall through.
2. Try the default-locale message catalog. If there's an ICUError, log the error and fall through.
3. Raise KeyError.

I think the simplest implementation is to rip out all the `Optional` from `MessageTranslator`. `MessageTranslator.get_message()` would raise `KeyError` if there's an error; `MessageTranslator.trans()` would raise `KeyError` or `ICUError`. And `localize()` would handle them all, just as `MessageTranslator.trans()` used to.

Also, would it be a good idea to name it `MessageTranslator.get_and_format_message()` instead of `MessageTranslator.trans()`? The name itself would make it obvious which errors might be raised.

giorgio93p: Part of this behaviour is new, in the sense that `trans` used its `default` argument as a last resort and we want to stop doing that. Hence, if the message is nowhere to be found, `localize` (and `trans`) raises the `KeyError` without handling it, correct?

server/serializers.py
giorgio93p: Well, a JSON-serialized representation would look really bad. On the other hand, it will contain the maximum amount of information possible for the user (at least if it does not get covered by other elements).

adamhooper: Meh, I'm not opinionated about JSON-serializing. It's just as wrong as anything else. As long as we log the error, don't crash, and _do_ include the message ID, we'll be fine.

cjworkbench/i18n/trans.py
adamhooper: Correct.

The `KeyError` is new, yes ... but that shouldn't affect `trans()` in practice. `trans()` can only raise `KeyError` if a developer fails to `bin/dev extract-messages` -- in which case, we _want_ a crash. Aside from that one case, I like the `trans()` behavior on master (its throwing of `ICUError`, for instance); so let's keep it.

In contrast, `localize()` is new. And the more I think about it, the more I feel it shouldn't handle any exceptions at all. Imagine:

```
def localize(locale_id: str, message_id: str, parameters: _Parameters) -> str:
    translations = _get_translations(locale_id)
    return translations.get_and_format_message(message_id, parameters)  # raise KeyError, ICUError
```

No need for default arguments. The only callers are `trans()` and `jsonize_i18n_message()`, and they both pass arguments. (Also, nit: this should be named "arguments", not "parameters", as ICU MessageFormat terminology is "arguments".)

No need for exception handlers. Callers handle exceptions differently.

No need for `Optional`. It's just added confusion.

It's easy to envision how `trans()` would work with this. And it's easy to envision how `jsonize_i18n_message()` would work.

Here's my proposal:

* `trans()` is a general-purpose function. It is used within Workbench to log Workbench messages. It crashes if the _developer_ made a mistake -- either forgot to extract messages or supplied an invalid message format.
* `localize()` is a low-level utility function.
* `jsonize_i18n_message()` is a high-level function specifically designed to handle I18nMessages from modules. Those messages are untrusted, so the tasks are A) log errors so module authors can see them; and B) recover from errors so output is always sane.

giorgio93p: The only thing missing from your comment is that `localize`, although low-level, has the job of implementing the fallback mechanism to the default locale. Hence, it needs to handle the errors of missing or badly formatted messages in the message catalogs of non-default locales.

For completeness, I should mention that with this pull request `trans` will also crash if the developer edits the English po file directly and makes a message invalid: it will crash when trying to use the message.

adamhooper: Well, in my previous comment I was suggesting a bit of a different direction. I was suggesting `localize()` should do less, `trans()` and `jsonize_i18n_message()` should pick up the slack. But that doesn't matter right now -- your new edits to this pull request cover all today's scenarios, so it's worth merging as-is. Later, when we implement multiple sources, we'll adjust.

giorgio93p:
@adamhooper By accident, commit https://github.com/CJWorkbench/cjworkbench/commit/ccfba1294239dd4ee5b332d967c16a3dc503b6e8 included some changes to `docker-compose.yml` and `docker-compose.commands.yml`

adamhooper:
I noticed that, and I forgot to slow things down a bit. Could you please add a commit that _comments_ on the need for `privileged: true`? The comment should describe which Linux distro is having problems and the debugging steps we've taken.

I'm willing to accept that those `privileged: true` are necessary. But they'll cost us later: we'll add features that will pass in unit tests and fail in integration tests, and it'll be frustrating to debug them. Consider our future selves when writing it. Our future selves will want to hear: "We understand you went through lots of debugging and ended up enraged at this `privileged: true` line. But you can't delete this line in a fit of rage. Your pain was a necessary part of the development process."

giorgio93p:
The thing is that I am not yet convinced that `privileged: true` is necessary, since 

1. only I have needed it and 
2. I haven't even managed to figure out why

adamhooper:
Okay, then please add a comment along the lines of `# TODO: nix this privileged: true\n# see https://github.com/CJWorkbench/cjworkbench/pull/174#issuecomment-578783281`

adamhooper:
(I agree we'd be better off without it. But I don't think it's a security concern on production, and I'd rather focus on i18n right now.)

PR #175, opened 2020-01-26
Write `.pot` file on module message extraction

giorgio93p:
Until now, modules did not get a `.pot` file. Pontoon does not like that.

In order to be able to create this `.pot` file even for modules that have no new messages, we needed a mechanism to force writing their catalogs even when there are no new messages. I opted for a flag in the `extract-module-messages` command.

By the way, while checking that `extract-messages` is still working, I found that two messages in workbench's catalogs were outdated. These are included here.


giorgio93p:
The intention is, as with workbench's po files, to have two files for English:

- `messages.po` is the one we actually use in `localize` and has the strings that we will actually see in the application
- `messages.pot` is the one that pontoon wants to see; it has empty strings and include the English strings as comments

It is assumed that the two files have the same message IDs and thus they are written together. The first time we try to extract strings from modules _whose strings have been extracted before_, this will be false, because one of them will be empty; this is where `force_update` comes into play.

adamhooper:
... I'm confused about two things here.

First, backtracking a bit: my understanding is that `.pot` files are language-agnostic. The only English in `.pot` files are comments. Other projects don't put `.pot` files in their `en/` directories. Shouldn't our files be in `locale/messages.pot`?

Second: I feel the decision on overwriting `messages.pot` should be unrelated to the decision on overwriting `messages.po`. Why do we tie the two together?

giorgio93p:
First: I think pontoon expects it inside `en`
Second: We want to keep them in sync

adamhooper:
First: that's putting the cart before the horse. We confused Pontoon to look inside `en/` _because_ that's where the file is. Not the other way around. We can certainly change the location of the `.pot` file in Pontoon.

Second: how would our messages become out of sync? ... and from a completely different angle, what is the purpose of `messages.pot` if it can't become out of sync with `messages.po`?

giorgio93p:
First: Looks like you are correct. By default, pontoon looks for the `.pot` file in one of these subfolders: 1) `templates`, 2) `en-US`, `en-us`, `en_US`, or 'en_us`, 3) `en`. Looks like there is also the ability to provide a custom path by including a `.toml` file in the repo. So we could use either a `templates` subfolder or a `.toml` file; your choice (the first seems easier)

Second: From my point of view, the only purpose of `messages.pot` is to keep pontoon happy. Actually, with or without `force_update`, this command will not make them out of sync; however, treating them together helps to make the intention clearer.

adamhooper:
I think I hear what you're saying: _we_ don't need potfiles, because we're perfectly happy to extract from scratch when we add a new language. That's fine. But we're learning that other gettext-based tools, such as Pontoon, need potfiles. So we might as well support potfiles.

Can we nix `force_update` and instead overwrite `messages.pot` if the `messages.pot` has changed (as opposed to if `messages.po` has changed)?

PR #176, opened 2020-01-27
Support for module localization in server.serializers

giorgio93p:
This pull request gives `jsonize_i18n_message` and `jsonize_clientside_module` the ability to search module catalogs.


cjworkbench/i18n/trans.py
```diff
@@ -90,22 +79,67 @@ def localize_html(
...
-        except KeyError as err:
-            pass
-    return _get_translations(default_locale).get_and_format_html_message(
-        message_id, context=context, arguments=arguments, tags=tags
+    return MessageLocalizer.for_application_messages(locale_id).localize_html(
+        message_id, arguments=arguments, tags=tags, context=context
     )
 
 
+class MessageLocalizer:
```
adamhooper: This class has two functions:

1. It has `.localize()` and `.localize_html()`.
2. Its class acts as a singleton registry of all messages.

Let's break it into two classes. The new class could be `MessageLocalizerRegistry` ... and then we'd have a singleton, `MESSAGE_LOCALIZER_REGISTRY`. Look to `cjwstate/models/module_registry` for inspiration.

cjworkbench/i18n/trans.py
```diff
@@ -90,22 +79,67 @@ def localize_html(
...
+    @lru_cache(maxsize=len(supported_locales))
+    def for_application_messages(cls, locale_id: str):
+        """Return a `MessageLocalizer` for the application internal messages in the given locale.
+
+        Uses lru_cache in order to parse the message catalogs only once per locale.
+        """
+        return cls(locale_id, load_catalog(locale_id), load_catalog(default_locale))
+
+    @classmethod
+    def for_module(cls, module_zip: ModuleZipfile, locale_id: str):
```
adamhooper: `module_zipfile`

cjworkbench/i18n/trans.py
```diff
@@ -90,22 +79,67 @@ def localize_html(
...
+        return cls(
```
adamhooper: We should cache here. For two reasons: A) it seems wrong to construct a new `MessageLocalizer` every time we call `for_module()`; and B) `cjwstate` shouldn't depend on Babel.

Let's use a class-member dictionary cache, keyed by module. `self._module_localizers: Dict[str, Dict[str, MessageLocalizer]] = weakref.WeakKeyDictionary()`. Use [WeakKeyDictionary](https://docs.python.org/3/library/weakref.html#weakref.WeakKeyDictionary): it won't hold a reference to the ModuleZipfile, and it'll delete catalogs belonging to deleted ModuleZipfiles. Remember to use a `threading.Lock` when accessing the dictionary.

cjworkbench/i18n/trans.py
adamhooper: `for_module_zipfile_and_locale_id()`

cjwstate/modules/types.py
```diff
@@ -13,6 +13,14 @@
 from cjwstate.modules.param_dtype import ParamDType
 from .module_loader import validate_module_spec
 from .param_spec import ParamSpec
+from cjworkbench.i18n import supported_locales
```
adamhooper: Do not import from `cjworkbench` into `cjwstate`. That's a circular dependency. (All our existing imports of this type are bugs.)

cjwstate/modules/types.py
```diff
@@ -13,6 +13,14 @@
...
+from babel.messages.pofile import read_po, PoFileError
```
adamhooper: Do not import Babel. Medium-term, `renderer` and `fetcher`, which use `cjwstate`, may be distributed without Babel.

cjwstate/modules/types.py
```diff
@@ -274,3 +282,17 @@ def get_param_schema_version(self) -> str:
             # "develop" is handled specially in cjwstate/params.py. For GitHub
             # sha1, a new version forces migrate_params().
             return self.version
+
+    @lru_cache(maxsize=len(supported_locales))
```
adamhooper: Please nix this and rename `_read_bytes()` to `read_bytes()`.

server/serializers.py
```diff
@@ -258,17 +505,58 @@ def jsonize_clientside_tab(tab: clientside.TabUpdate) -> Dict[str, Any]:
     return d
 
 
+def _localize_module_message(
+    message_id: str,
+    *,
+    default: Optional[str],
+    ctx: JsonizeModuleContext,
+    arguments={},
+    log_not_found=False,
```
adamhooper: Why would we _not_ want to log when a message isn't found?

server/serializers.py
```diff
@@ -258,17 +505,58 @@ def jsonize_clientside_tab(tab: clientside.TabUpdate) -> Dict[str, Any]:
     return d
 
 
+def _localize_module_message(
+    message_id: str,
+    *,
+    default: Optional[str],
```
adamhooper: Nix `default`. Where does it come from?

The answer is: "it comes when we're translating the specfile." But there's a better way to do that:

`def _localize_module_spec_message(message_path: str, ctx: JsonizeModuleContext, specfile_message: str)`

... and then simplify _this_ method by making it accept an `I18nMessage`.

server/serializers.py
```diff
@@ -68,7 +71,29 @@ def _camelize_value(v: Any) -> Any:
         return v
 
 
-JsonizeContext = namedtuple("JsonizeContext", ["user", "session", "locale_id"])
+JsonizeContext = namedtuple(
+    "JsonizeContext", ["user", "session", "locale_id", "localizers"]
```
adamhooper: Do we really need to add `localizers` here? Isn't there already a cache on `MessageLocalizer.for_application_messages()` that makes this redundant?

server/serializers.py
```diff
@@ -68,7 +71,29 @@ def _camelize_value(v: Any) -> Any:
...
+)
+JsonizeModuleContext = namedtuple(
+    "JsonizeModuleContext", ["user", "session", "locale_id", "module_id", "localizers"]
```
adamhooper: Again, can we avoid `localizers`?

And if we can't, let's make them `application_localizer`, `module_localizer`. But I think we can.

server/serializers.py
```diff
@@ -68,7 +71,29 @@ def _camelize_value(v: Any) -> Any:
...
+        locale_id=ctx.locale_id,
+        module_id=module_id,
+        localizers=ctx.localizers,
+    )
+
+
+def collect_i18n_message_sources(locale_id: str):
+    return {
+        f"module.{module_id}": MessageLocalizer.for_module(module_zip, locale_id)
+        for module_id, module_zip in MODULE_REGISTRY.all_latest().items()
```
adamhooper: You cannot call `MODULE_REGISTRY.all_latest()` here, because it hits the database from the event-loop thread. I'll add a comment to `ModuleRegistry` warning against this pattern.

Nix `collect_i18n_message_sources()`. It is unusable. And we don't need to -- a cache in `MessageLocalizer` should be faster, anyway.

server/tests/test_serializers.py
```diff
@@ -1,17 +1,1421 @@
...
+            module_id=module_id,
+            localizers=localizers,
+        )
+    else:
+        return JsonizeContext(
+            user=user, session=session, locale_id=locale_id, localizers=localizers
+        )
+
+
+def _jsonize_clientside_module_return(args: Dict[str, Any]) -> Dict[str, Any]:
```
adamhooper: This should be a variable, not a function. Call it `DEFAULT_SERIALIZED_MODULE = { ... }`

And then callers should compare with `{**DEFAULT_SERIALIZED_MODULE, ...}`

server/tests/test_serializers.py
```diff
@@ -1,17 +1,1421 @@
...
+    result.update(args)
+    return result
+
+
+class JsonizeClientsideModuleTest(unittest.TestCase):
+    def test_with_js(self):
+        module = Module(
+            ModuleSpec(
+                **{
+                    "id_name": "testme",
```
adamhooper: While you're editing here, could you please nix the `**{` pattern? I was being lazy when I put it in.

This should look like `ModuleSpec(id_name="testme", name="Test Module", category="Clean", parameters=[])`. Ditto everywhere in this file.

server/tests/test_serializers.py
```diff
@@ -1,17 +1,1421 @@
...
+        result = jsonize_clientside_module(
+            module,
+            mock_jsonize_context(
+                locale_id="el",
+                localizers={
+                    "module.testme": MessageLocalizer("el", catalog, default_catalog)
+                },
+            ),
+        )
+        expected = _jsonize_clientside_module_return(
```
adamhooper: `self.assertEqual(result["name"], "Translated name")`

If we compare the whole dict, the reader can't tell what's under test. We should have maybe one test that checks the entire structure; other than that, we should test the things we're testing, not the whole dict.

server/tests/test_serializers.py
```diff
@@ -1,17 +1,1421 @@
...
+        self.assertDictEqual(result, expected)
 
 
 class JsonizeI18nMessageTest(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        MessageLocalizer.for_application_messages.cache_clear()
+
+    def tearDown(self):
+        super().tearDown()
```
adamhooper: Move `super().tearDown()` to be the last line of `tearDown()`.

server/serializers.py
giorgio93p: Because not finding the translation of a module spec is normal at this point given the current structure of the rest of the code. The module author may not have opted to use i18n, so the (pybabel) catalog this function accesses (via a `MessageLocalizer`) will not contain the corresponding key.

giorgio93p: Now that I think of it, `True` would be a better default for this argument

server/serializers.py
giorgio93p: Although it's more meaningful when used for translating the specfile, the `default` argument also covers the case of `jsonize_i18n_message`, which has its own default value as per [this comment](https://github.com/CJWorkbench/cjworkbench/pull/174#discussion_r369620491).

server/serializers.py
adamhooper: There's a difference between "The translation is missing" and "The module isn't translated".

We _always_ want to log if the translation is missing. We _never_ want to log if the module isn't translated. This function has all the information it needs to make this decision (assuming `ModuleLocalizer.for_module_zipfile_and_locale_id()` raises `KeyError`).

server/serializers.py
adamhooper: Please nix it anyway. The function itself has all the information it needs to generate it.

cjwstate/modules/types.py
giorgio93p: Nix caching of the whole `get_message_catalog` method? I think it's better to keep `get_message_catalog`, in order to define the catalog path in there.

adamhooper: Yes, nix caching.

You're right that it's a good idea to define the path. Since it won't return a catalog, shall we call it ... uh ... `read_messages_po_for_locale()`? (That name doesn't roll off the tongue, but it's important that we describe the bytestream format, because the return type will simply be `bytes`.)

server/serializers.py
giorgio93p: The question then is who will call `MODULE_REGISTRY.all_latest()` (via something like `MESSAGE_LOCALIZER_REGISTRY.update_modules()`).

Since `MODULE_REGISTRY` actually changes in only a few circumtances, we could initialize `MESSAGE_LOCALIZER_REGISTRY` on system startup and only run `MESSAGE_LOCALIZER_REGISTRY.update_modules()` when doing stuff that changes modules. I tried doing this (among other stuff...) in commits https://github.com/CJWorkbench/cjworkbench/pull/176/commits/441eaf54cf4be4afe5119d03c7d9c36061064ac0 and https://github.com/CJWorkbench/cjworkbench/pull/176/commits/6a242d22446de1ca781b8bf925101d882655eeec; however, adding `MESSAGE_LOCALIZER_REGISTRY.update_modules()` to asgi creation broke the setting up of `server.tests.test_websockets`.

adamhooper: I don't understand the problem. `MODULE_REGISTRY.all_latest()` is already called. The `clientside.Module` is already passed to `jsonize_clientside_update()`.

adamhooper: Ah, I see the problem. Wanna chat about it on Slack?

server/serializers.py
```diff
@@ -492,48 +498,66 @@ def jsonize_clientside_tab(tab: clientside.TabUpdate) -> Dict[str, Any]:
...
-        return default
+    if not spec_value:
+        return spec_value
 
     try:
-        localizer = MESSAGE_LOCALIZER_REGISTRY.for_module_id(ctx.module_id)
-    except KeyError:
-        logger.exception(f"Unsupported module as I18nMessage source: {ctx.module_id}")
-        return default
+        return _localize_module_message(
```
adamhooper: Please don't implement this in terms of `_localize_module_message()`. Currently, `_localize_module_message()` is complex (it logs _some_ stuff but not _other_ stuff; it raises errors when it shouldn't) just to support reuse in this function. That's not a worthwhile compromise.

These two functions should have two different docstrings and two different sets of behavior. Each possible problem should be enumerated in the docstring and code.

server/views/embed.py
```diff
@@ -30,7 +30,12 @@ def embed(request, wfmodule_id):
                 wf_module = None
 
     if wf_module:
-        ctx = JsonizeContext(request.user, request.session, request.locale_id)
+        ctx = JsonizeContext(
+            request.user,
+            request.session,
+            request.locale_id,
+            dict(MODULE_REGISTRY.all_latest()),
```
adamhooper: No need for the registry. Just use `{module_zipfile.module_id: module_zipfile}`.

server/views/importfromgithub.py
```diff
@@ -18,8 +18,12 @@ def import_from_github(request):
 
     try:
         clientside_module = import_module_from_url(request.data["url"])
```
adamhooper: Please make `import_module_from_url()` return a `ModuleZipfile`. Then, instead of hitting the module registry, pass `{module_zipfile.module_id: module_zipfile}`

server/views/workflows.py
```diff
@@ -65,7 +65,12 @@ def make_init_state(
     except Workflow.DoesNotExist:
         raise Http404("Workflow was recently deleted")
 
-    ctx = JsonizeContext(request.user, request.session, request.locale_id)
+    ctx = JsonizeContext(
+        request.user,
+        request.session,
+        request.locale_id,
+        dict(MODULE_REGISTRY.all_latest()),
```
adamhooper: `modules`

server/views/workflows.py
```diff
@@ -82,7 +87,12 @@ def post(self, request: HttpRequest):
     def get(self, request: HttpRequest):
         """Render workflow-list page."""
 
-        ctx = JsonizeContext(request.user, request.session, request.locale_id)
+        ctx = JsonizeContext(
+            request.user,
+            request.session,
+            request.locale_id,
+            dict(MODULE_REGISTRY.all_latest()),
```
adamhooper: `{}`. There are no steps or messages -- no need for modules.

cjworkbench/i18n/trans.py
```diff
@@ -192,31 +200,53 @@ def for_module_zipfile(cls, module_zipfile: ModuleZipfile):
...
 
-    def for_module_id(self, module_id: str) -> MessageLocalizer:
-        with self._module_localizers_lock:
-            return self._module_localizers[self._supported_modules[module_id]]
+    def for_module_zipfile(self, module_zipfile: ModuleZipfile) -> MessageLocalizer:
+        """Return a `MessageLocalizer` for the given `ModuleZipFile`
 
-    def for_application(self):
-        return self._app_localizer
+        Raise `NotInternationalizedError` if the `ModuleZipFile` has no po files
```
adamhooper: What does it raise on invalid pofiles?

cjworkbench/i18n/trans.py
```diff
@@ -124,6 +128,10 @@ def for_module_zipfile(cls, module_zipfile: ModuleZipfile):
                     f"Invalid po file for module {module_zipfile.module_id_and_version} in locale {locale_id}: {err}"
                 )
                 catalogs[locale_id] = Catalog()
+            except KeyError as err:
+                pass
```
adamhooper: Why is this `pass`, while `PoFileError` leads to `Catalog()`?

I'd expect the two cases to lead to the same result (though only `PoFileError` would cause logging).

cjworkbench/i18n/trans.py
```diff
@@ -192,31 +200,53 @@ def for_module_zipfile(cls, module_zipfile: ModuleZipfile):
...
+                if isinstance(result, NotInternationalizedError):
+                    raise result
+                else:
+                    return result
+            except KeyError:
+                pass  # we'll fetch it now
+
+            # 3. Update the cache, still holding the lock.
+            try:
+                localizer = MessageLocalizer.for_module_zipfile(module_zipfile)
```
adamhooper: Please nix `MessageCatalogRegistry.for_module_zipfile()` and `MessageCatalog.for_module_zipfile()` and ... well ... _any_ other `.for_module_zipfile()` ... and use a private method on this class.

`MessageLocalizerRegistry` is a factory. It should not use factory methods from other classes.

cjworkbench/i18n/trans.py
```diff
@@ -248,6 +214,27 @@ def for_application(self) -> MessageLocalizer:
...
+                logger.exception(
+                    f"Invalid po file for module {module_zipfile.module_id_and_version} in locale {locale_id}: {err}"
+                )
+                catalogs[locale_id] = Catalog()
+            except KeyError as err:
+                pass
+        if not catalogs:
+            raise NotInternationalizedError(module_zipfile.module_id)
+        return MessageLocalizer(MessageCatalogsRegistry(catalogs))
+
```
adamhooper: Can we remove `MessageCatalogsRegistry`? It's _only_ used in `MessageLocalizer`, and as far as I can tell it's one method and a dict.

giorgio93p: So I move the dict and the method back to `MessageLocalizer`?

server/serializers.py
```diff
@@ -535,77 +539,89 @@ def _localize_module_spec_message(
...
         return spec_value
 
 
+def _i18n_message_source_to_localizer(
+    message: I18nMessage, ctx: JsonizeContext
+) -> MessageLocalizer:
+    """Return a localizer for the source of the given `I18nMessage`.
+    
+    Raise `NotInternationalizedError` if the source is a non-internationalized module.
+    Raise `JsonizeContextError` if the source is a module that has no associated `ModuleZipFile`.
```
adamhooper: Raise `KeyError` if the query is for a module ID that is is not in `ctx`.

PR #178, opened 2020-02-04
Support i18n messages from cjwmodule

giorgio93p:
While we are at it, simplify I18nMessage.source


cjworkbench/i18n/trans.py
```diff
@@ -209,6 +212,34 @@ def for_application(self) -> MessageLocalizer:
         """Return a `MessageLocalizer` for the application messages"""
         return self._app_localizer
 
+    @lru_cache(1)
+    def for_cjwmodule(self) -> MessageLocalizer:
+        """Return a `MessageLocalizer` for the messages of `cjwmodule`"""
+        catalogs = {}
```
adamhooper: I would expect `for_cjwmodule()` and `for_application()` to share design patterns, since their behavior is so very similar. For instance, if `for_application` uses `self._app_localizer()`, then `for_cjwmodule()` should use `self._cjwmodule_localizer()`.

Please nix `cjworkbench/i18n/catalogs/__init__.py` and move its functions into here. Nix `BadCatalogsError` -- everywhere it occurs, we actually want a crash. And nix `UnsupportedLocaleError` because it wraps an `Exception` and I'd rather we have the underlying exception.

To clarify, here is the behavior we should be having here:

* If `application` is missing a pofile even though we mark it a supported locale, _crash_. (Simply ... don't catch anything. I'd expect a `FileNotFoundError`; if it's something else, maybe add a comment to that effect.)
* If `cjwmodule` is missing a pofile, log a warning and supply an empty catalog, so translations will use the default locale. (This will help us support new languages out-of-order -- i.e., translate `cjworkbench` before translating `cjwmodule`.)
* If `application` _or_ `cjwmodule` contains an invalid pofile, _crash_. (Don't catch anything.)

giorgio93p: Some of the stuff in `cjworkbench/i18n/catalogs/__init__.py` is also used in scripts under `cjworkbench/i18n/catalogs`

cjwkernel/types.py
```diff
@@ -596,7 +596,7 @@ class I18nMessage:
     """Where the message comes from (`"module"` or `"cjwmodule"`), or `None` if it comes from Workbench proper."""
 
     def __post_init__(self):
-        if self.source is not None and self.source not in ["module", "cjwmodule"]:
+        if self.source not in [None, "module", "cjwmodule"]:
```
adamhooper: `%r` below, not `%s`. `%s` can crash.

cjworkbench/tests/i18n/util.py
```diff
@@ -7,15 +7,14 @@
...
-    # Code to acquire resource, e.g.:
-    old_localizer = MESSAGE_LOCALIZER_REGISTRY._app_localizer
-    try:
-        new_localizer = MessageLocalizer(catalogs)
-        MESSAGE_LOCALIZER_REGISTRY._app_localizer = new_localizer
-        yield new_localizer
-    finally:
-        # Code to release resource:
-        MESSAGE_LOCALIZER_REGISTRY._app_localizer = old_localizer
+    def mock_for_application(registry):
```
adamhooper: You don't need `@contextmanager` to decorate `patch()` -- you can just return from `patch()` and that will be a context manager.

And if you're using `patch()`, then consider `patch.object()` -- it's less finicky. (Your usage here is correct in this case, but `patch.object()` is a better pattern in general.)

I propose:

```
def mock_app_catalogs(catalogs: Dict[str, Catalog]) -> ContextManager:
    return patch.object(MODULE_LOCALIZER_REGISTRY, "application_localizer", MessageLocalizer(catalogs))
```

cjworkbench/i18n/trans.py
```diff
@@ -208,9 +204,13 @@ def for_module_zipfile(self, module_zipfile: ModuleZipfile) -> MessageLocalizer:
             # Release the lock. If another caller is waiting for us to release
             # the lock, now it should check self._cache again.
 
+    @lru_cache(1)
```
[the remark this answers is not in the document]
giorgio93p: OK. This implies that until we add (dummy) message catalogs to `cjwmodule`, the constructor will always raise an exception while trying to find them.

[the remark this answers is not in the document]
adamhooper: Oh yeah. Thank you for reminding me.

I'll update `cjwmodule` right now. It'll take 1-2 hours for Docker Hub to finish with it. Then you can merge master and we can do this right.

adamhooper:
@giorgio93p Once you fix the unit-test failures, this will be good to go :)

giorgio93p:
I can't see where the CI tests failed in the last commit. Everything seems to pass when I run it.

adamhooper:
It might have been an intermittent integration-test failure. (An element reference is "stale".) I'm re-running the tests; if they pass, I'll merge.

PR #182, opened 2020-02-10
Add module id to clientside.Update when the update contains I18nMessages

giorgio93p:
Issue: with module id missing from `clientside.Update`, there is no way to tell in which module's catalogs to search, hence `server.serializers.jsonize_clientside_step` adds `None` as module id to the jsonize context which causes to `server.serializers._i18n_message_source_to_localizer` raise `KeyError`.

### DIFF
--- a/cjwstate/clientside.py
+++ b/cjwstate/clientside.py
@@ -163,7 +163,11 @@ class StepUpdate:
     """
     Module identifier.
 
-    Required for creation, and must be None afterwards.
+    Required for creation, and must usually be None afterwards.
+    
+    When a `render_result` is provided, you must also pass a `module_slug`.
+    This is so that `I18nMessage`s with source `"module"` in the result
+    can be interpreted relative to their source module.
     """
 
     tab_slug: Optional[str] = None
--- a/cjwstate/clientside.py
+++ b/cjwstate/clientside.py
@@ -163,11 +163,8 @@ class StepUpdate:
     """
     Module identifier.
 
-    Required for creation, and must usually be None afterwards.
-    
-    When a `render_result` is provided, you must also pass a `module_slug`.
-    This is so that `I18nMessage`s with source `"module"` in the result
-    can be interpreted relative to their source module.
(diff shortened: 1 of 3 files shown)


cjwstate/clientside.py
```diff
@@ -163,7 +163,8 @@ class StepUpdate:
     """
     Module identifier.
 
-    Required for creation, and must be None afterwards.
+    Required for creation, and must usually be None afterwards,
```
adamhooper: Required for creation and whenever `render_result` is non-empty.

cjwstate/clientside.py
```diff
@@ -196,6 +197,10 @@ class StepUpdate:
     If the cached-result delta ID is different from the _relevant_ delta ID,
     then the result is obsolete. The client may render obsolete data plus a
     progress indicator.
+    
+    When a `render_result` is provided, you must also pass a `module_slug`.
+    This is so that `I18nMessage`s with source `"module"` in the result
+    can be interpreted relative to their source module.
```
adamhooper: Please add a `__post_init__()` to assert this is the case. Otherwise we can easily regress. (As you've seen, the error is tricky to spot and to debug.)

`if self.render_result is not None: assert self.module_slug is not None`

renderer/execute/wf_module.py
```diff
@@ -447,7 +447,14 @@ def _execute_wfmodule_save(
     crr, output_delta = await _execute_wfmodule_save(workflow, wf_module, result)
 
     update = clientside.Update(
-        steps={wf_module.id: clientside.StepUpdate(render_result=crr)}
+        steps={
+            wf_module.id: clientside.StepUpdate(
+                render_result=crr,
+                # We also add `module_slug`, in order for `I18nMessage`s with source `"module"`
```
adamhooper: No comment here, thanks. `StepUpdate` docs are perfectly clear.

adamhooper:
To speed things up, I'm going to merge this and do the changes myself. Thank you!

PR #226, opened 2021-01-04
Download data as csv 175651036

pierreconti:


adamhooper:
@pierreconti ping!

adamhooper:
Closing. We're calling it "Download data as CSV", and it's already deployed to prod.

PR #259, opened 2021-07-09
Change text to Report URL

pierreconti:


adamhooper:
@pierreconti This looks like a glossary change. Shall we make the site consistent?

As a user viewing a workflow, I click "Report" and I see my report. Why is there a "Report URL" component? I would expect the report URL to be in the address bar. What's the difference between the URL in my address bar (since _I'm on the report page_) and the "Report URL" (which, presumably, _is the report page_.)

We're on familiar turf here for users. Workbench's Report feature is basically a CMS, and tons of users have experience with CMSs. They all look like:

* Wordpress puts a "Post URL" in its admin interface. Its editor pages are called "Edit Post".
* Medium calls its page "Story" and its editor pages "Edit Story".

So I'm all for calling the report page "Report" ... _if_ we call the other page "Edit Report" or "Report Editor".

pierreconti:
Yes agree, I was dwelling on that change.

PR #302, opened 2023-10-22
Remove links to retired server from README

tfmorris:
Per this [tweet](https://twitter.com/workbenchdata/status/1483597535486742529), the public server was shutdown a couple of years ago, so leaving a link to it in the README is a just an attractive nuisance. My browser plugins block loading the page, so I suspect it's either owner by domain squatters or some other nefarious types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,6 @@
 
 Welcome to Workbench!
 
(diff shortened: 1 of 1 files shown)


adamhooper:
Thank you!

PR #303, opened 2024-12-03
fix(docker-compose): Remove duplicate staticfiles service entry

hydrosquall:
## Motivation

Per the [wiki](https://github.com/CJWorkbench/cjworkbench/wiki/Setting-up-a-development-environment#developer-workflow), `bin/dev start`  was not launch as the docker-compose file was not parseable by the yaml library.

image

## Changes

Removes duplicate section from `docker-compose.yml` (removed section matches)

https://github.com/CJWorkbench/cjworkbench/blob/0f7433139ba45d7c379d677f93aebb1c6c419168/docker-compose.yml#L128-L151

## Testing

I'm not able to get the app running locally for unrelated reasons (pandas, pybind11 , google-re2, numpy are not installing), but I'll work on debugging those separately. We won't be able to reach that step unless we clear the YAML installation step first.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,7 +75,7 @@ services:
       context: .
       target: jsbase
     volumes:
-      # Migrate 
+      # Migrate
       - type: bind
         source: ./
         target: /app/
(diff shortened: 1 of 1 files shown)


adamhooper:
Thank you! Awesome :)

